PR #9, opened 2024-06-13
Import upstream fixes duckdb/extension-ci-tools#31

carlopi:
Fix to Linux, allowing retry in case of bad network, and avoid hardcoding Windows toolchain.


samansmink (approved):
LGTM

PR #12, opened 2024-06-14
Changes to CI, build by default on all platforms, but allow overrides

carlopi:
Allow extension to declare which platform are not supported, and otherwise build for all platforms.

Also change CI slightly so at least 1 extensions is always build (as a sanity check)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,7 @@ jobs:
       - prepare
       - build
     uses: ./.github/workflows/_extension_deploy.yml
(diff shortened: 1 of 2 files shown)


carlopi:
[This](https://github.com/carlopi/community-extensions/actions/runs/9518126439) is a run, on my fork, where `wasm_mvp` is disabled for sqlite_scanner, to check that the logic actually works.

carlopi:
Working as expected, for me it can go.

samansmink:
LGTM!

PR #14, opened 2024-06-26
Add PRQL extension

ywelsch:
The PRQL extension (see https://github.com/ywelsch/duckdb-prql) is based on an older copy of the extension template. It might need to be brought up-to-date first for this to work?


ywelsch:
The PRQL extension needs the PRQL transpiler, which is written in Rust. The build is currently failing because the Rust toolchain is absent in this build environment.

samansmink:
@ywelsch ah got ya, we will need to get that in first, i can take a stab at that. It's basically all there in the delta extension

rustyconover:
I have a fix in my repo for the linux_arm64 cross compile.

carlopi:
@ywelsch (& also @rustyconover, given you are here and the comment is the same): I think on main I have the fix for the documentation test failing.

Could you merge or rebase on latest main? Thanks! Then extensions should pass.

---
Independent, we can also change this later, I think in both case it would be nice to have it explicit from the descriptor that this is a C++ + Rust extension. I think probably the simpler is having the language as `C++ & Rust`, would that make sense? That would allow also to enable branching on including Rust toolchain or not (see comment here: https://github.com/duckdb/community-extensions/pull/22#pullrequestreview-2154125483)

samansmink:
I think we can just merge this first, fix the docs thing later

carlopi:
Actually, moving forward, congrats @ywelsch!

ywelsch:
I've just pushed another update :)

samansmink:
okay lets just let it run then merge both PRs

PR #18, opened 2024-06-30
Add shellfs

rustyconover:


extensions/shellfs/description.yml
```diff
@@ -0,0 +1,13 @@
+extension:
+  name: shellfs
+  description: Allow shell commands to be used for input and output
+  version: 1.0.0
+  language: C++
+  build: cmake
```
carlopi, line 6:
```suggestion
  build: cmake
  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads"
```

I think for the Web-case the concept of spawning shell task has little meaning, and in the case of server side while it could be possible make some sense, it would still escape the Wasm sandbox.
I think it's safer, to get started, to avoid packaging for those.

PR #21, opened 2024-07-02
fix: update Readme with logo and add description

rustyconover:


carlopi:
Thanks! Picture is funny, but we were thinking of something more boring.

rustyconover:
🦆 👍🏻

PR #22, opened 2024-07-02
Add support for rust-based extensions

samansmink:
By switching to the workflow from extension-ci-tools, we can now support extensions that rely on the rust toolchain to be present

This *should* unblock https://github.com/duckdb/community-extensions/pull/19 and https://github.com/duckdb/community-extensions/pull/14


.github/workflows/build.yml
```diff
@@ -40,14 +40,15 @@ jobs:
 
   build:
     needs: prepare
-    uses: ./.github/workflows/_extension_distribution.yml
+    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.0.0
     if: ${{ needs.prepare.outputs.COMMUNITY_EXTENSION_NAME != '' }}
     with:
       duckdb_version: v1.0.0
+      enable_rust: true
```
carlopi, line 47:
```suggestion
      enable_rust: ${{ needs.prepare.outputs.COMMUNITY_EXTENSION_ENABLE_RUST }}
```
Possibly?

carlopi: This would need also the connected change in the script that pass down the environment variables.

carlopi (approved):
Left a note on always enabling rust, I think it's fine as-is, but given it's just matter of passing down a flag we might even considering having it dependent on a given extension's descriptor.

I checked the diff between old file and what will be introduced (via extension-ci-tools), looks good!

samansmink:
yea good point, I will add that in a follow up PR, lets push forwards to unblock the other 2 PRs first!

PR #26, opened 2024-07-03
Improve docs markdown generation, allow example and extended 

carlopi:
Add possibility for extension to have two extra fields in the descriptor file:
* docs.hello_world: an optional SQL based example of what capabilities the extension provides
* docs.extension_description: allow to pass an optional extra paragraph (markdown enabled) to describe / provide additional context on the extension

That allows to generate markdown files, that once passed down to the website generation outputs something like:
image

HTML side improvements have to be made, but those are orthogonal on deciding what content should be there.

### DIFF
--- a/extensions/quack/description.yml
+++ b/extensions/quack/description.yml
@@ -11,3 +11,9 @@ extension:
 repo:
   github: hannes/quack
   ref: 09680a975bcf9e93b5a2f46d3eeb68792d5239c6
+
+docs:
+  hello_world: |
+    SELECT quack(i::VARCHAR) FROM range(10) tbl(i);
+  extended_description: |
+    The quack extension is based on DuckDB's [Extension Template](https://duckdb/extension_template/), and it's a great starting point to get started building more advanced extensions.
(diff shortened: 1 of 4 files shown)


extensions/quack/description.yml
```diff
@@ -10,4 +10,10 @@ extension:
 
 repo:
   github: hannes/quack
-  ref: 09680a975bcf9e93b5a2f46d3eeb68792d5239c6
\ No newline at end of file
+  ref: 09680a975bcf9e93b5a2f46d3eeb68792d5239c6
+
+docs:
+  hello_world: |
+    SELECT quack(i::VARCHAR) FROM range(10) tbl(i);
```
samansmink, line 17: nit: maybe this can actually quack hello world at the user?
```suggestion
    SELECT quack('world');
```

this outputs :
```
┌────────────────┐
│ quack('world') │
│    varchar     │
├────────────────┤
│ Quack world 🐥 │
└────────────────┘
```
which seems cute for a hello world example

samansmink (approved):

PR #31, opened 2024-07-05
fix: add evalexpr_rhai

rustyconover:


carlopi (commented):
Thanks!

Linux extension CI needs updating on our side, so that's on us.

Windows CI is supposed to work though, could you take a look at the failure ? It has something to do with _impl_free / _impl_malloc from what I got from the stack trace.

rustyconover:
Hi @carlopi thanks for the review, I see the error on Windows too.

I'll look into it tomorrow.

carlopi:
One thing to consider, you can consider skipping Windows CI until that gets sorted out.
See this for reference: https://github.com/duckdb/community-extensions/pull/32/files (you would need likely to skip both windows_amd64 and windows_amd64_rtools)

carlopi (changes requested):
One thing to consider, you can consider skipping Windows CI until that gets sorted out.
See this for reference: https://github.com/duckdb/community-extensions/pull/32/files (you would need likely to skip both windows_amd64 and windows_amd64_rtools)

carlopi (approved):

PR #40, opened 2024-07-09
fix: improve readme

rustyconover:


carlopi:
Thanks, here there are a couple of items on our side, I will need to take a look on how to automate this a bit more (triggering CI on README is weird / how to have README table autogenerated).

PR #43, opened 2024-07-09
Add chsql extension

lmangani:
Adding chsql extension to Community 🤞 

> This extension provides a growing number of Clickhouse SQL Macros for DuckDB. 

Special thanks @carlopi for all the support and assistance backporting the builders to v1.0.0 ⭐

### DIFF
--- a/extensions/chsql/description.yml
+++ b/extensions/chsql/description.yml
@@ -1,0 +1,19 @@
+extension:
+  name: chsql
+  description: Clickhouse SQL Macros for DuckDB
+  version: 1.0.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - lmangani
+
+repo:
+  github: lmangani/duckdb-extension-clickhouse-sql
+  ref: 21c240ec0cbef00d5928083f5fc936ec42542911
+
+docs:
+  hello_world: |
+    SELECT toString('world') as hello, toInt8OrZero('world') as zero;
+  extended_description: |
+    This extension provides a growing number of Clickhouse SQL Macros for DuckDB. 
--- a/extensions/chsql/description.yml
+++ b/extensions/chsql/description.yml
@@ -2,7 +2,7 @@ extension:
   name: chsql
   description: Clickhouse SQL Macros for DuckDB
   version: 1.0.0
-  language: C++
+  language: SQL & C++
   build: cmake
   license: MIT
   maintainers:


extensions/chsql/description.yml
```diff
@@ -0,0 +1,19 @@
+extension:
+  name: chsql
+  description: Clickhouse SQL Macros for DuckDB
+  version: 1.0.0
+  language: C++
```
carlopi, line 5:
```suggestion
  language: SQL & C++
```

carlopi (approved):
I gave a brief looks, we would need to write a proper template for SQL-only extensions with @Alex-Monahan, but that's for later.

I noticed in the macros there is a:
```
CREATE OR REPLACE MACRO intDiv(a, b) AS (a / b);
CREATE OR REPLACE MACRO intDivOrZero(a, b) AS COALESCE((a / b), 0);
```
but `//` should be used for integer division, otherwise you will get a floating point number as result.

Thanks, and Welcome!

lmangani:
> I gave a brief looks, we would need to write a proper template for SQL-only extensions with @Alex-Monahan, but that's for later.

Is anything out of place? I'd be curious to learn and improve and/or act as a guinea duck anytime

> CREATE OR REPLACE MACRO intDiv(a, b) AS (a / b);
> 
> but `//` should be used for integer division, otherwise you will get a floating point number as result.

We don't actually use the aliases sql file, its purely a reference for testing and understanding the logic.

The [actual code](https://github.com/lmangani/duckdb-extension-clickhouse-sql/blob/main/src/chsql_extension.cpp#L71
) is as follows: 
```
{DEFAULT_SCHEMA, "intDiv", {"a", "b"}, R"((CAST(a AS BIGINT) / CAST(b AS BIGINT)))"},
```

Does it look proper?

lmangani:
Exciting and works out of the box 😂 
image

carlopi:
I think you still need a double `//`:
```
D INSTALL chsql FROM community;
D LOAD chsql;
D SELECT intDiv(5,2);
┌──────────────┐
│ intdiv(5, 2) │
│    double    │
├──────────────┤
│          2.5 │
└──────────────┘
```
see also https://duckdb.org/docs/sql/functions/numeric.html#division-and-modulo-operators
```
D SELECT 5/2;
┌─────────┐
│ (5 / 2) │
│ double  │
├─────────┤
│     2.5 │
└─────────┘
D SELECT 5//2;
┌──────────┐
│ (5 // 2) │
│  int32   │
├──────────┤
│        2 │
└──────────┘
```

lmangani:
@carlopi totally missed that and wrongly assumed you were referring to casting. 
Great input! I will update the macro and experiment with our first extension upgrade :)

lmangani:
Fixed in 17c249ba08a9b88338e77c7f2d6e5dd2040b4590 
@carlopi should I just repoint the extension ref or is a minor version increase required to cycle?

carlopi:
Truth is that version field is, at the moment, not really used, only relevant information, also used on extension install is the ref.
In the run-up to 1.1 we will iterate and specify more precisely the meaning of the version.

PR #48, opened 2024-07-11
Adding DuckPGQ extension

Dtenwolde:
Hi everyone,

This PR introduces the [DuckPGQ](https://github.com/cwida/duckpgq-extension) extension, which supports the SQL/PGQ syntax for more succinct pattern-matching and path-finding syntax in graph queries. I have included some initial examples in the documentation and plan to expand on these in the future.

The extension is unique in its build process. It doesn't build against duckdb/duckdb but rather against a fork, [cwida/duckdb-pgq](https://github.com/cwida/duckdb-pgq). Apart from the parser (lib_pg_parser) and transformer, this fork is binary-compatible with DuckDB. This compatibility allows the extension to be loaded into any DuckDB instance like any other extension. Here’s how you can do it (after launching with the `unsigned` flag):

```bash
set custom_extension_repository = 'http://duckpgq.s3.eu-north-1.amazonaws.com';
force install 'duckpgq';
load 'duckpgq';
```

This method has been functional from version 0.10.1 up to and including 1.0.0.

@carlopi has noted that Go/Rust might face issues due to building with link-time optimizations, which I will investigate further. However, I don't believe this is a blocker for this PR.

I understand that this approach is somewhat unconventional for the extension framework. However, after discussing with @carlopi, we decided to submit this PR to have a discussion and gather feedback.

Looking forward to your thoughts and feedback!


carlopi (approved):
Thanks, and Welcome!

PR #50, opened 2024-07-12
add requires_toolchains options

samansmink:
This option allows specifying optional extra toolchains that are required for building. So far we have only the rust toolchain. But we would like to be able to add more.

it can be specified as a `;` separated list of toolchain names, that are passed straight down into `duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml`.

I've enabled the option for the existing extensions that depend on rust, and removed the now deprecated `enable_rust` option. This means that every extension added to the community extensions that needs rust, will need to specify it in its description file

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,6 @@ jobs:
     if: ${{ needs.prepare.outputs.COMMUNITY_EXTENSION_NAME != '' }}
     with:
       duckdb_version: v1.0.0
-      enable_rust: true
       exclude_archs: ${{ needs.prepare.outputs.COMMUNITY_EXTENSION_EXCLUDE_PLATFORMS }}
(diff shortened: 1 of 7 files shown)


carlopi (approved):
Thanks! Current infra I think requires max 1 descriptor change per PR, but unsure if that needs fixing today.

Looks good to me, I think we can also merge + then spawn CI (via explicit trigger) in one of the affected extensions (say PRQL) just to double-check.

samansmink:
I've launched a CI run with only the PRQL one touched here https://github.com/samansmink/community-extensions/actions/runs/9908263501

I would suggest to wait for that to pass then merge this

samansmink:
Ci in fork passes, so since we can't currently run multiple extensions at once we will just merge this

PR #53, opened 2024-07-13
Update chsql to v1.0.2

lmangani:
Updating to latest version with 100+ ClickHouse macros supported (native + aliases)


lmangani:
Thanks for merging @carlopi 
Tiny side question: how often are the community extension docs updated? Is it a manual process?
https://community-extensions.duckdb.org/extensions

carlopi:
Currently there is some caching in the middle that make so that, depending on cache status, website and extensions might be up to 1 day or so older. We are getting that sorted out, but it's not there yet.

Doc generation & deployment is done by a script, but currently is triggered manually, but should be made so it's scheduled, also that we know how to but has not been rolled out yet.

PR #62, opened 2024-07-16
Add workflow to deploy website

carlopi:
Script takes current state of duckdb/community-extensions combined with the latest state of the `community-extension` branch in duckdb-web.

Workflow is to be triggered, either manually or from another worfklow.

Calling mechanism is not yet there, first allowing to iterate on this.


carlopi:
Example of this in action is, on my fork, https://github.com/carlopi/community-extensions/actions/runs/9953957364/job/27498572303.

This fails due to missing privileges, but apart from that it should demonstrate this works.

PR #67, opened 2024-07-17
Fix csv list generation and avoid unnecessary rebuilds

carlopi:
Fix script that automatically populates `http://community-extensions.duckdb.org/extensions/community_extensions.csv` (cached version is off due to incompatibilities between linux and macos behaviour of the script).

Also avoid rebuild extension pipeline if no relevant changes are detected.

Next up: changes to docs.hello_world or docs.extended_description should not trigger rebuild.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,17 @@
 name: Community Extension Build
 on:
   pull_request:
+    paths-ignore:
+      - '**'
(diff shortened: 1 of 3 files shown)


carlopi:
Tested on my fork that CSV after this changes looks like:
[community_extensions.csv](https://github.com/user-attachments/files/16266979/community_extensions.csv)

PR #72, opened 2024-08-16
Update README.md

samansmink:
We don't want to maintain this I think, the community extensions website should contain the auto-updated list of community extensions


carlopi:
LGTM!

PR #81, opened 2024-08-28
Add workflow to build and test all current extensions vs a given duckdb version

carlopi:
This would allow to ramp up testing against current main, in the run-up to duckdb v1.1.0.

Example of cumulative workflow: https://github.com/carlopi/community-extensions/actions/runs/10603679975.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,14 @@ on:
     inputs:
       extension_name:
         type: string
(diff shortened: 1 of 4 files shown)


carlopi:
CI here is checking that build still keeps functional.

Real value is added in the additional invokable workflow that builds all current extensions versus 3 architectures (that have been chosen since they are faster than others but cover some variations in setup / machines).

PR #84, opened 2024-08-29
Add fuzzycomplete extension

rustyconover:


extensions/fuzzycomplete/description.yml
```diff
@@ -0,0 +1,100 @@
...
+  version: 1.0.0
+  language: C++
+  build: cmake
+  license: Apache-2.0
+  requires_toolchains: "rust"
+  maintainers:
+    - rustyconover
+
+repo:
+  github: https://github.com/rustyconover/duckdb-fuzzycomplete-extension
```
carlopi, line 13:
```suggestion
  github: rustyconover/duckdb-fuzzycomplete-extension
```

rustyconover:
Thanks for the fix @carlopi

carlopi:
There looks to be a problem in the extension init:
```
Invalid Input Error: Initialization function "fuzzycomplete_init" from file "build/extensions/v1.0.0/linux_amd64_gcc4/fuzzycomplete.duckdb_extension" threw an exception: "Table Function with name "sql_auto_complete" already exists!"
```
see the docs_test run, that actually tests loading the extension.

carlopi:
This also suggest we should add such a test as part of regular CI on all platforms

rustyconover:
Hey @carlopi,

The autocomplete extension and this extension provide the same function that is called by the CLI.  The error is likely because autocomplete is loaded or built in and when this extension is loaded there is a conflict.  Can you change the doc building step not to load the autocomplete extension?

Thanks,

Rusty

carlopi:
`autocomplete` does comes statically linked in some widely used duckdb clients, for example the brew-distributed CLI, so I don't think it's viable to remove that (at this moment in the development process).

I do think the test does make sense to be performed in a real word scenario, but that's me.

I would think this could also be fixed at the extension level, possibly allowing registration of the function only if `sql_auto_complete` is not already registered, or possibly overriding the current function (it that doable? I guess so) or some similar mechanism within the extension.

Would that make sense to you?

carlopi:
For example changing the LoadInternal function from:
```c++
		TableFunction auto_complete_fun("sql_auto_complete", {LogicalType::VARCHAR}, SQLFuzzyCompleteFunction,
																		SQLFuzzyCompleteBind, SQLFuzzyCompleteInit);
		ExtensionUtil::RegisterFunction(db, auto_complete_fun);
```
to
```c++
                if (!db.ExtensionIsLoaded("autocomplete")) {
	 	    TableFunction auto_complete_fun("sql_auto_complete", {LogicalType::VARCHAR}, SQLFuzzyCompleteFunction,
																		SQLFuzzyCompleteBind, SQLFuzzyCompleteInit);
		    ExtensionUtil::RegisterFunction(db, auto_complete_fun);
                }
```

carlopi:
Thanks!

PR #86, opened 2024-08-30
fix: add ref_next for lindel

rustyconover:
For the upcoming DuckDB 1.1.0 release there were changes to calls to LogicalType::ARRAY, these changes live on a branch and add a reference to the latest commit to that branch.


carlopi:
Great, thanks

PR #88, opened 2024-08-31
Add example usage for faiss extension

JAicewizard:
Add example for the faiss extension, as a comment really isn't that great and the linked readme doesn't have an example either

### DIFF
--- a/extensions/faiss/description.yml
+++ b/extensions/faiss/description.yml
@@ -19,6 +19,19 @@ repo:
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


extensions/faiss/description.yml
```diff
@@ -19,6 +19,19 @@ repo:
 
 docs:
   hello_world: |
-    -- github.com/arjenpdevries/faiss/blob/main/README.md
+    load faiss;
```
carlopi, line 22:
```suggestion
```
This should not be needed in the example, it's already included by default

PR #99, opened 2024-09-09
chsql 1.0.3 for 1.1.0

lmangani:
Let's go 1.1.0 🎉


lmangani:
@carlopi not sure what can be dropped from the backport version. Any chance you could hint me in the right direction?

carlopi:
Thanks!

PR #100, opened 2024-09-09
fix: change git ref for duckdb-lindel-extension

rustyconover:


carlopi:
Thanks! Sorry for the noise.

PR #101, opened 2024-09-09
fix: change git ref for duckdb-crypto-extension

rustyconover:


rustyconover:
@carlopi can you check this one out? Thank you very much.

carlopi:
Thanks!

PR #108, opened 2024-09-10
chsql extension docs 

lmangani:
Adding docs csv for chsql - including a function_type column since they're mostly macros - is any column allowed/required?
The format is based on assumptions from other extensions and needs a little validation from someone who knows best 🙏


carlopi:
Thanks!

carlopi:
Result will be something like: 
image

lmangani:
Thanks for the confirmation @carlopi next we'll make this part of each release as we add macros 🎉 

BTW is there any actual documentation for the extension details formats? I'd be happy to contribute to a little guide if useful.

PR #111, opened 2024-09-16
pivot_table extension for spreadsheet style pivots

Alex-Monahan:
Open to feedback! Not sure how to test this ahead of time, so please let me know how I should test. It compiles locally at least!

The goal is to publish a blog post about this on 2024-09-20, but there is an open slot in the schedule for the following week as well. 

@szarnyasg as an FYI since this is also something the blog post will depend on!

### DIFF
--- a/extensions/pivot_table/description.yml
+++ b/extensions/pivot_table/description.yml
@@ -1,0 +1,19 @@
+extension:
+  name: pivot_table
+  description: Provides a spreadsheet-style pivot_table function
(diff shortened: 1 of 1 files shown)


carlopi (approved):
Thanks!!

carlopi:
I think this is ready to go, I will later give this a spin.

Alex-Monahan:
NICE! It works! Thanks Carlo!!

PR #112, opened 2024-09-16
Add xxhash extension written in Go with the C API. 

ajzo90:


ajzo90:
Having trouble with building the docs. Complains about missing .so file. 
`IO Error: Extension "build/extensions/v1.1.0/linux_amd64_gcc4/xxhash.duckdb_extension" could not be loaded: libduckdb.so: cannot open shared object file: No such file or directory`

Worth mentioning is that I'm using the new C API extensions, perhaps related to that?

https://github.com/ajzo90/duckdb-community-extensions/actions/runs/10886389437/job/30207261474

carlopi:
What's failing is loading the extension in DuckDB itself, I guess locally works since you have the `libduckdb.so` or equivalent in the right path, but that is not the case for the environment where the docs are generated.

I think you do need to link in libduckdb in your extension explicitly to make this work.

ajzo90:
Giving up on this for now, succeeded building the extension with statically linking duckdb, but got stuck on another error instead.

ajzo90:
It is now stuck in 
```Run ./duckdb -c "SET extension_directory = 'build/extension_dir'; FORCE INSTALL 'build/downloaded/$NAME.duckdb_extension';"
IO Error: Extension "build/downloaded/xxhash.duckdb_extension" could not be loaded: build/downloaded/xxhash.duckdb_extension: cannot allocate memory in static TLS block`
```

carlopi:
Hi, I have to say I am not exactly what's happening here, it might about some linking flag or so, unsure.

I understand @taniabogatsch is taking a look on how to create go extensions with the C API, so help might be coming!

taniabogatsch:
I haven't looked into building Go extensions with the current extension template.
But indeed, I am looking into building them with the new extension C API.

PR #117, opened 2024-09-22
Get latest pivot_table to always order by all

Alex-Monahan:
Should fix flaky tests. CC @carlopi !


Alex-Monahan:
Looks like my `limit` trick did not solve the ordering. I'm going to need to get a bit more creative. I'll work on it.

Alex-Monahan:
This is working now! Ready to merge I believe.

PR #120, opened 2024-09-24
fix: change git ref for duckdb-lindel-extension

rustyconover:


carlopi:
Hi! I think this is ready to be merged, other PR's CI I have currently suspended, and will restart / merge them later, to make room for the beefy job that builds all extensions with current defaults: https://github.com/duckdb/community-extensions/actions/runs/11014272173. Thanks for the understading.

PR #123, opened 2024-09-24
fix: change git ref for duckdb-shellfs-extension

rustyconover:


carlopi:
@rustyconover: problem with the failing test is that we changed build system, and now python3 is not available anymore (or not guaranteed to be there).

I guess this can be "fixed" either extension side, basically either temporarily skipping the test or moving it to be about something else, or it can be fixed more properly tomorrow on our side.

PR #124, opened 2024-09-24
fix: change git ref for duckdb-crypto-extension

rustyconover:


samansmink:
Hey @rustyconover! I don't actually think these PRs  that only bump your submodule should be necessary. As longs as your extension builds on the main branch of DuckDB, we will just rebuild all community extensions targeting the new duckdb release.

I sent you this on discord as well, but for future reference:

for next (v1.2.0) duckdb release we will:
- some time (probably a few weeks) before the planned release date we will try to build all community extension using the main DuckDB branch
- we open an issue in all community extension repo's that are failing for main: in this issue we ask each extension dev to fix their builds and set the `ref_next` field to the fixed commit/version in the community extension descriptor file
- Now the CI in https://github.com/duckdb/community-extensions should have everything we need to build the extensions for v1.2.0 before we actually make the v1.2.0 tag in duckdb. (using the mechanism from https://github.com/duckdb/extension-ci-tools/pull/83) 

This way all community extensions that either build as-is, or that specify a functioning `ref_next` during the duckdb release, will be pre-built and immediately available when v1.2.0 is released

PR #126, opened 2024-09-25
Update DuckPGQ for v1.1.1

Dtenwolde:


Dtenwolde:
I'm not exactly sure where it goes wrong, but it currently builds against duckdb/duckdb, while the extension is supposed to build against cwida/duckdb-pgq. Is there something I missed setting up? My own CI passed https://github.com/cwida/duckpgq-extension/actions/runs/11040549104. Please let me know :)

carlopi:
I will have a look later, CI setup changed in https://github.com/duckdb/extension-ci-tools/pull/79, unsure of that or elsewhere something is now off.

carlopi:
@Dtenwolde: I think the problem is here: https://github.com/cwida/duckpgq-extension/blob/f2204a248ff6524f71e262d4c36b71c57444caee/src/include/duckpgq/core/utils/compressed_sparse_row.hpp#L18, where it should likely work with `""` instead of `<>`.

Your build system is also wired differently, using duckdbpgq instead of duckdb as base folder, so problems are not visible there.

PR #151, opened 2024-10-16
Format example queries in community extension descriptors

szarnyasg:
This PR applies some basic SQL formatting on the example queries. Chief among these is the consistent use of capitalization of SQL keywords: while many capitalize `SELECT`, they do not do so for `FROM`, `AS`, `JOIN`, etc.


carlopi:
Thanks a lot!

carlopi (approved):

PR #153, opened 2024-10-16
chsql: Update link

szarnyasg:


carlopi:
Thanks, this needs to go in after the other is live.

PR #159, opened 2024-10-21
Flockmtl

queryproc:
This PR adds `FlockMTL`, a DuckDB extension, as a community supported extension. 

`FlockMTL` introduce `PROMPT`(s) and `MODEL`(s) as DDL constructs and multiple LLM scalar functions, namely `llm_complete`, `llm_complete_json`, `llm_filter`, and `llm_embedding`.

The current ref points to the following release: https://github.com/dsg-polymtl/duckdb-flockmtl/releases/tag/v0.1.0

### DIFF
--- a/extensions/flockmtl/description.yml
+++ b/extensions/flockmtl/description.yml
@@ -13,10 +13,12 @@ extension:
 
 repo:
   github: dsg-polymtl/flockmtl
-  ref: f46ba8f9c2538e994373ead0f103859316048682
+  ref: 86e32ae7ac5ea4743067d2ddae9a96c7d3a2cd23
 
 docs:
   hello_world: |
+    -- The library loads successfully if an OPENAI_API_KEY is set in the environment
+
     -- Call an OpenAI model with a predefined prompt ('Tell me hello world') and default model ('gpt-4o-mini')
     D SELECT llm_complete('hello-world', 'default');
     ┌──────────────────────────────────────────┐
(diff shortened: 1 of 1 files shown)


queryproc:
Doing: Fixing windows

EDIT: fixed the issue! (Will bump the ref and @ you once ready for the build)

anasdorbani:
@samansmink can you pls run the builds the extension is ready?

carlopi (commented):
Thanks!

Currently this fails one test for generating docs, since the assumption is that extensions can be loaded without other dependencies, but there a missing env variable makes so the extension can't load.

I will have a look at making doc generation step take a setup step, so that a fake env variable can for example be added. This might make sense also in other cases to set invariants or so. Or maybe make so that they can be skipped. I will come back on this.

queryproc:
> Thanks!
> 
> Currently this fails one test for generating docs, since the assumption is that extensions can be loaded without other dependencies, but there a missing env variable makes so the extension can't load.
> 
> I will have a look at making doc generation step take a setup step, so that a fake env variable can for example be added. This might make sense also in other cases to set invariants or so. Or maybe make so that they can be skipped. I will come back on this.

@carlopi Okay. Thank you! Will load a garbage env var value for now to be able to pass and let the extension fail on function use.

queryproc:
@carlopi added the fix to skip the env var loading. We can run the workflow!

p.s. might be worth adding this test to the original GH action extension template.

PR #163, opened 2024-10-23
Add pretty-printed star count and download count numbers

szarnyasg:
I added pretty-printed fields to the Markdown generator, with the rationale being that it's a bit more convenient to do it here than to do it from Jekyll/Liquid.


carlopi:
```sql
CREATE MACRO pretty_number(x) AS (
    CASE
      WHEN x < 1000 THEN x || ''
      WHEN x < 1000000 THEN ((x/1000)::DECIMAL(10,1)) || 'K'
      ELSE 'a lot'
    END);
```
The need for this macro in DuckDB is closer and closer

carlopi (approved):
Thanks!

carlopi:
(bash script do not really makes sense, but will fix it while rebasing)

carlopi:
Moving to draft, this will not properly work if the either stars of download numbers are empty / null. I can fix it later.

szarnyasg:
@carlopi I fixed the issue and improved the ergonomics of the scripts. PTAL.

scripts/generate_md.sh
```diff
@@ -82,11 +82,18 @@ do
...
-       echo "" >> $EXTENSION_README
-       echo -n "extension_download_count: " >> $EXTENSION_README
-       cat build/downloads-last-week.json | jq ".${extension}" >> $EXTENSION_README
+
+       STAR_COUNT=$(python3 scripts/get_stars.py extensions/$extension/description.yml $1)
+       STAR_COUNT_PRETTY=$(echo $STAR_COUNT | python3 scripts/pretty_print.py)
+       echo "extension_star_count: $STAR_COUNT" >> $EXTENSION_README
+       echo "extension_star_count_pretty: $STAR_COUNT_PRETTY" >> $EXTENSION_README
+
+       DOWNLOAD_COUNT=$(cat downloads-last-week.json | jq ".${extension}")
```
carlopi, line 91:
```suggestion
       DOWNLOAD_COUNT=$(cat build/downloads-last-week.json | jq ".${extension}")
```

scripts/generate_md.sh
```diff
@@ -82,11 +82,18 @@ do
...
+       DOWNLOAD_COUNT_PRETTY=$(echo $DOWNLOAD_COUNT | python3 scripts/pretty_print.py)
+       echo "extension_download_count: $DOWNLOAD_COUNT" >> $EXTENSION_README
+       echo "extension_download_count_pretty: $DOWNLOAD_COUNT_PRETTY" >> $EXTENSION_README
+
+       cat build/downloads-last-week.json | jq ".$extension" >> $EXTENSION_README
```
carlopi, line 96:
```suggestion
```

carlopi (changes requested):

carlopi (approved):

carlopi:
Looks great, thanks!

Updating the markdowns is blocked on extensions downloads numbers being emtpy...

PR #167, opened 2024-10-28
New Extension: HostFS

gropaul:
### HostFS Extension 

This small extension enables navigating and exploring the host computer's file system. I created it because I was often unsure of a specific file's name I wanted to open when working in the DuckDB shell. Also, by navigating to the directory data, you can open a file without needing to enter (and know) the long absolute path. 

#### SQL Example
```sql 
-- Navigate to the workspace and list the files
D PRAGMA cd('/Users/paul/workspace');
D PRAGMA ls;
┌───────────────────────────────┐
│             path              │
│            varchar            │
├───────────────────────────────┤
│ ./duckdb                      │
│ ./playground                  │
│ ./duckfs                      │
-- Find the files you were working on last
D SELECT path, file_last_modified(path) AS date FROM ls() WHERE 'csv' IN file_extension(path) ORDER BY date LIMIT 1 ;
┌───────────────────────────┬─────────────────────┐
│           path            │        date         │
│          varchar          │      timestamp      │
├───────────────────────────┼─────────────────────┤
│ ./sketch_results_join.csv │ 2024-07-13 23:25:48 │
└───────────────────────────┴─────────────────────┘
-- List the top 3 file types by total size, with file count, ordered by size.
D SELECT size, count, file_extension AS "type"
FROM (
SELECT SUM(file_size(path)) AS size_raw, hsize(size_raw) AS size, COUNT(*) AS count, file_extension(path) AS file_extension
FROM lsr('/Users/paul/workspace', 10)
GROUP BY file_extension(path)
) AS subquery
ORDER BY size_raw DESC LIMIT 3;
┌───────────┬───────┬─────────┐
│   size    │ count │  type   │
│  varchar  │ int64 │ varchar │
├───────────┼───────┼─────────┤
│ 246.95 GB │    29 │ .duckdb │
│ 90.33 GB  │  3776 │ .tmp    │
│ 26.17 GB  │ 28175 │ .csv    │
└───────────┴───────┴─────────┘
```

### DIFF
--- a/extensions/duckfs/description.yml
+++ b/extensions/duckfs/description.yml
@@ -1,0 +1,53 @@
+extension:
+  name: duckfs
+  description: Navigate and explore the filesystem using SQL
+  version: 0.0.1
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - Gropaul
+
+repo:
+  github: gropaul/duckFS
+  ref: 83f66a15a216baa1ab9b7348663dde4de1db2ba0
+
+docs:
+  hello_world: |
+    -- Navigate to the workspace and list the files
+    D PRAGMA cd('/Users/paul/workspace');
+    D PRAGMA ls;
+    ┌───────────────────────────────┐
+    │             path              │
+    │            varchar            │
+    ├───────────────────────────────┤
+    │ ./duckdb                      │
+    │ ./playground                  │
+    │ ./duckfs                      │
+    -- Find the files you were working on last
+    D SELECT path, file_last_modified(path) AS date FROM ls() WHERE 'csv' IN file_extension(path) ORDER BY date LIMIT 1 ;
+    ┌───────────────────────────┬─────────────────────┐
+    │           path            │        date         │
+    │          varchar          │      timestamp      │
+    ├───────────────────────────┼─────────────────────┤
+    │ ./sketch_results_join.csv │ 2024-07-13 23:25:48 │
+    └───────────────────────────┴─────────────────────┘
+    -- List the top 3 file types by total size, with file count, ordered by size.
+    D SELECT size, count, file_extension AS "type"
+    FROM (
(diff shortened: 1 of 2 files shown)


carlopi:
This looks good to me, I assume this will fail on docs creation, I need to look at this, but it's not on your side.

carlopi (approved):

gropaul:
Thank you 😊

carlopi:
@gropaul: I think this is ready, only doubt I have is whether the name is not off, since this is NOT using DuckDB own file system (as far as I can tell), so for example it will not work only locally.

We don't have a proper naming convention in place, so it's likely OK to be merged. @samansmink, unsure if you have more formalized thoughts.

samansmink (approved):
formalized thoughts, me? Ha 😁 

Jokes aside, I think @carlopi's point is definitely valid, though. It is probably an important gotcha to realize that this does not go through duckdb's own filesystem which could be a potential footgun.

Additionally, some of the functions in this extension seems to be already available in main duckdb, e.g. https://duckdb.org/docs/sql/functions/char.html#format_bytesbytes and https://duckdb.org/docs/sql/functions/char.html#parse_dirnamepath-separator

Nevertheless, I don't think these are blockers for this PR!

gropaul:
I am very free with the name and did not think about DuckDB's on FS, good point! Do you think _SystemFS_ / _HostFS_ would be a more appropriate name?

Also, thanks for pointing out the existing format functions, I will remove the duplicates from the extension.

samansmink:
Well i guess it depends on your plan for the extension right? if this will only ever be intended to directly interface with the host filesystem through c++, perhaps HostFS does make sense! If you plan to include other filesystems too perhaps the current one is fine?

Maybe you could also start with HostFS then rename later if the functionality is expanded to go through duckdb's own filesystem? I think the rename would be justified in that case as it really does change fundamentally

carlopi:
Welcome!

PR #172, opened 2024-11-03
GSheets v0.0.2

archiewood:
Updates DuckDB GSheets to v0.0.2
- Makes OAuth default provider
- Fixes bugs with reading
- Support for writing all types


archiewood:
Not sure why, but the docs do not seem to have updated

carlopi:
Should be up

PR #178, opened 2024-11-05
Update DuckPGQ for v1.1.3

Dtenwolde:
Bumping DuckPGQ to 0.1.0 to mark that property graphs are now persistent. 
This version is compatible with DuckDB v1.1.3. 
Currently running the CI/CD for this commit here https://github.com/cwida/duckpgq-extension/actions/runs/11683847628 (Awaiting final checks at the time of writing)


carlopi:
Thanks!

PR #185, opened 2024-11-09
Update http_client to v0.0.3

lmangani:
### What's Changed
- Add new http_get extra function with header & params by @ahuarte47 in https://github.com/quackscience/duckdb-extension-httpclient/pull/8
- Excluded `windows_amd64_mingw` from targets due to incompatibilities we'll address if there's demand

### DIFF
--- a/extensions/http_client/description.yml
+++ b/extensions/http_client/description.yml
@@ -1,26 +1,33 @@
 extension:
   name: http_client
   description: DuckDB HTTP Client Extension
-  version: 0.0.2
+  version: 0.0.3
   language: C++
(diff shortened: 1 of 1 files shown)


lmangani:
Set to review ready assuming the action errors are unrelated to the code/release & happy to update/sync anytime

carlopi (approved):
Thanks!

After merging https://github.com/duckdb/community-extensions/pull/186, CI should be good here

lmangani:
Is this due to the exclusion we added? In any case, no worries it can wait!

PR #214, opened 2024-12-01
Enforce ordering by function name and type

szarnyasg:
Applying `ORDER BY function_name` does not warrant a deterministic ordering as the same function name can have multiple function types (table/scalar). This can lead to oscillation as seen here https://github.com/duckdb/duckdb-web/pull/4223/files.

This PR enforces deterministic ordering by adding the function type in the ordering criteria.


carlopi (approved):
Thanks!!

PR #225, opened 2024-12-09
Add tsid extension

lmangani:


carlopi:
Hi! This seems to me somehow similar to https://duckdb.org/community_extensions/extensions/ulid from @Maxxen, unsure if you were aware.

Obv there is space for everyone, but wanted to point to connected work in the area.

lmangani:
@carlopi its similar indeed (as most *IDs in general) but ULID seems barely documented and I'm not sure what the plan is. Our TSID extension uses a larger size ID and exposes functionality to extract timestamps and seed generation. I guess they can co-exist?

PR #234, opened 2024-12-15
new flockmtl release v0.2.1

queryproc:
Update the description to point to the latest release: https://github.com/dsg-polymtl/flockmtl/releases/tag/v0.2.1

### DIFF
--- a/extensions/flockmtl/description.yml
+++ b/extensions/flockmtl/description.yml
@@ -1,6 +1,6 @@
 extension:
   name: flockmtl
-  description: DuckDB LLM & RAG Extension
+  description: DuckDB LLM & RAG Extension [(Docs)](https://dsg-polymtl.github.io/flockmtl/docs).
   version: 0.2.0
   language: SQL & C++
   build: cmake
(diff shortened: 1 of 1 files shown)


extensions/flockmtl/description.yml
```diff
@@ -1,6 +1,6 @@
 extension:
   name: flockmtl
-  description: DuckDB LLM & RAG Extension
+  description: DuckDB LLM & RAG Extension [(Docs)](https://dsg-polymtl.github.io/flockmtl/docs).
   version: 0.2.0
```
carlopi, line 4:
```suggestion
  version: 0.2.1
```

carlopi: I think this should be 0.2.1, I will assume so and commit with this, but please @queryproc check on this

carlopi:
Thanks, left a comment, I think version field should be 0.2.1, but good to go anyway.

extensions/flockmtl/description.yml
queryproc: Yeah, you are right. I will update the version with the next release.

PR #252, opened 2025-01-06
Bump avro next

hannes:


extensions/avro/description.yml
```diff
@@ -12,7 +12,8 @@ extension:
 
 repo:
   github: hannes/duckdb_avro
-  ref: e5ed59b6ccf915c65e17eb6286b9a64f3ab09f59
+  ref: 7facc0badf31c7ec0a249cf47fb97d190550d3f2
```
carlopi, line 15:
```suggestion
  ref: e5ed59b6ccf915c65e17eb6286b9a64f3ab09f59
```

carlopi:
Adapted `ref`, and in the meantime I enabled CI on ref_next on any extension, so should test both that `ref` paired with `v1.1.3` work AND that `ref_next` paired with current main duckdb branch (to eventually become v1.2.0 work).

PR #258, opened 2025-01-08
Skip `_mingw` whenever skipping `_rtools`

carlopi:
Platform selection changed a bit between `v1.1.3` and the upcoming `v1.2.0`:

* `windows_amd64_rtools` has been removed, replaced by `windows_amd64_mingw`
* `linux_amd64_musl` has been added

The first change is relevant here, where whenever `_rools` is mentioned `_mingw` is added.

This should have no real impact, unless you are downloading extenisions manually or builing a duckdb-client. Details are in https://github.com/duckdb/duckdb/pull/14368.

### DIFF
--- a/extensions/avro/description.yml
+++ b/extensions/avro/description.yml
@@ -5,7 +5,7 @@ extension:
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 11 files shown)


carlopi:
Note that failure is expected, given more that 1 extension per PR has been changed, but for cleanup PRs such as these it should still be OK.

PR #265, opened 2025-01-10
[shellfs] Add ref_next for branch supporting 1.2

rustyconover:
Add a ref_next for shellfs with changes to support upcoming 1.2 release.


rustyconover:
@carlopi any ideas what to do about this failure?

carlopi:
@rustyconover: I will investigate, this is connected to renaming (that is equivalent of removal + adding a new one) of a platform.

Not a blocker anyhow, thanks

PR #269, opened 2025-01-13
[lindel] fix: bump version

rustyconover:
There was a bug with Hilbert and Morton decoding as reported at https://github.com/rustyconover/duckdb-lindel-extension/issues/4

I've fixed that bug and added a test case.  Bumping the extension to version 1.0.1.


rustyconover:
@carlopi any hint about the failure for deploy on `windows_rtools`?  Should we ignore it?

carlopi:
@rustyconover: let me double check what's happening here, then I will merge the PR.

PR #272, opened 2025-01-15
Bump h3 to get fixes for functions accepting lists of H3 indexes

isaacbrodsky:
Applies https://github.com/isaacbrodsky/h3-duckdb/pull/141

Does this need a new version number / release or is that independent of the ref?


carlopi:
I think it should be good to go, thanks!

PR #274, opened 2025-01-19
Add zipfs

isaacbrodsky:


carlopi:
It seems to be compatible with main (great), but not with v1.1.3.

I think either:
* waiting for v1.2.0, and then merging
* add some way to say "ref: -; ref_next: abcdef012345;", where `-` means skip this
* fixing at the code level (unsure if worth)

isaacbrodsky:
> It seems to be compatible with main (great), but not with v1.1.3.
> 
> I think either:
> 
>     * waiting for v1.2.0, and then merging
> 
>     * add some way to say "ref: -; ref_next: abcdef012345;", where `-` means skip this
> 
>     * fixing at the code level (unsure if worth)

I initially developed it against v1.1.3, so I think that is just a mistaken edit when I was updating for `main`. I updated the ref once again, and I think it will turn green now...

carlopi:
Thanks a lot!

PR #276, opened 2025-01-21
remove windows_amd64_rtools build

queryproc:


extensions/flockmtl/description.yml
```diff
@@ -5,12 +5,10 @@ extension:
   language: SQL & C++
   build: cmake
   license: MIT
-  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads"
+  excluded_platforms: "windows_amd64_rtools;wasm_mvp;wasm_eh;wasm_threads"
 
   maintainers:
     - dorbanianas
-    - SunnyYasser
-    - queryproc
```
samansmink, line 13: Just checkin @queryproc, Is this on purpose?

queryproc: Yes. Thanks for checking!

PR #285, opened 2025-02-07
Create the first release (v1.2.0) of the Substrait extension

EpsilonPrime:
The Substrait extension previously was a DuckDB core extension.  This is the first release of the Substrait extension as a community extension.

### DIFF
--- a/extensions/substrait/description.yml
+++ b/extensions/substrait/description.yml
@@ -9,7 +9,7 @@ extension:
     - anshuldata
     - cgkiran
     - EpsilonPrime
-  excluded_platforms: "windows_amd64_rtools;windows_amd64_mingw"
+  excluded_platforms: "windows_amd64_rtools;windows_amd64_mingw;windows_amd64"
 
 repo:
   github: substrait-io/duckdb-substrait-extension
--- a/extensions/substrait/description.yml
+++ b/extensions/substrait/description.yml
@@ -1,0 +1,23 @@
+extension:
+  name: substrait
+  description: Allows conversion execution of Substrait query plans
+  version: 1.2.0
+  language: C++
+  build: cmake
+  license: Apache-2.0
+  maintainers:
+    - anshuldata
+    - cgkiran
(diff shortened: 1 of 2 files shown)


carlopi:
Failure mode looks like to be about external C++ tests, if there is something missing on the community-extensions side, an issue (or a PR) to make them work in general is welcome. 

You might consider disabling when invoked via community-extensions, for the purpose of distributing the extensions.

EpsilonPrime:
Thanks, I'm looking into it.  I suspect the issue is that the extension isn't getting loaded.  There's probably some small but important difference between the extension's build and this one.

EpsilonPrime:
Looked at some other extensions -- looks like we aren't using extension-ci-tools which could be the culprit.

extensions/substrait/description.yml
```diff
@@ -0,0 +1,23 @@
...
+  description: Allows conversion execution of Substrait query plans
+  version: 1.2.0
+  language: C++
+  build: cmake
+  license: Apache-2.0
+  maintainers:
+    - anshuldata
+    - cgkiran
+    - EpsilonPrime
+  excluded_platforms: "windows_amd64_rtools;windows_amd64_mingw"
```
rymurr, line 12:
```suggestion
  excluded_platforms: "windows_amd64_rtools;windows_amd64_mingw;wasm_mvp;wasm_eh;wasm_threads"
```

EpsilonPrime: Thanks, the wasm builds worked so I left those in.  The windows_amd64 build failed most likely since the tests aren't made to work with Microsoft's C++ offering.

rymurr:
@carlopi would you be able to retrigger? I think my changes means the workflows aren't runnign without approval again

carlopi:
Thanks and welcome!

carlopi:
Builds now https://github.com/duckdb/community-extensions/actions/runs/13247994714, once that completes it will be available.

PR #286, opened 2025-02-07
[fuzzycomplete] Skip musl platform

carlopi:
There is some problem to be fixed with propagating rust toolchain, skipping to keep distributing elsewhere.

@rustyconover fyi

### DIFF
--- a/extensions/fuzzycomplete/description.yml
+++ b/extensions/fuzzycomplete/description.yml
@@ -171,6 +171,7 @@ extension:
   name: fuzzycomplete
   requires_toolchains: rust
   version: 1.0.0
+  excluded_platforms: "linux_amd64_musl"
 repo:
   github: rustyconover/duckdb-fuzzycomplete-extension
   ref: 39a61c1c39a2b028101ef635fcb04856883ab498


rustyconover:
This is a strange bug, I opened.

https://github.com/corrosion-rs/corrosion/issues/606

carlopi:
Thanks, It's very possibly we are not passing the right flags while cross-compiling. There is some work done in the rust extensions, that might solve also the problem here.

carlopi:
I though removing the extra platform was better to allow extension to be published, that should still be better.
Feel free to open a bug to duckdb/extension-ci-tools if you think the problem is our side.

rustyconover:
I’ve spent the morning trying to fix this by updating the `duckdb_extension.Makefile`.

I added this section to set the appropriate flag:

```makefile
### For Linux musl, set the proper cargo target
ifeq ("${DUCKDB_PLATFORM}", "linux_amd64_musl")
	RUST_FLAGS=-DRust_CARGO_TARGET=x86_64-unknown-linux-musl
endif
```

The flag gets set, but unfortunately, the build still fails because Corrosion seems to insist on having the `gnu` toolchain installed.

I'm wondering—how critical is the `musl` target? It feels like all Rust extensions might currently face the same issue. Perhaps they haven’t kept up with the latest changes in `extension-ci-tools` that introduced the `musl` target, I'll work on building this today.

PR #302, opened 2025-02-20
update chsql_native for v1.2.0

lmangani:
Updated chsql_native rust extension for v1.2.0

### Notes
* windows support excluded due to build issues. TBD if there's any demand.


carlopi:
Thanks

PR #304, opened 2025-02-21
fix: Sync the default vcpkg commit match the commit specified extension-ci-tools for 1.2.0

rustyconover:
If an extension didn't specify a `vcpkg` commit, the default was being used, but the default was not changed to match the default specified in `extension-ci-tools` for 1.2.0.


rustyconover:
@samansmink and @carlopi I expect the Windows to fail, because the CI check is running before the change is applied.

carlopi:
Thanks!

carlopi:
(the windows CI failed also since `quack` hardcodes the old commit, I will send a PR to fix that, edit: https://github.com/duckdb/community-extensions/pull/305)

PR #314, opened 2025-03-10
updated bigquery to v0.2.1

hafenkran:
Quick question: Just realized that the BigQuery extension isn't available in the community extension repository. Has the automatic build been disabled? Or were there any problems with the BigQuery extensions?

### DIFF
--- a/extensions/bigquery/description.yml
+++ b/extensions/bigquery/description.yml
@@ -5,7 +5,7 @@ extension:
   language: C++
   build: cmake
   license: MIT
-  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64;windows_amd64_rtools;windows_amd64_mingw;osx_amd64;linux_arm64"
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64_mingw;osx_amd64;linux_arm64"
   vcpkg_commit: "e01906b2ba7e645a76ee021a19de616edc98d29f"
   requires_toolchains: "parser_tools"
   maintainers:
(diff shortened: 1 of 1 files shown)


hafenkran:
The build is passing for me while it breaks here. I was able to trace down this issue to this line [here](https://github.com/duckdb/community-extensions/blob/717aa62a9f69f6183349775884d150cc1332df79/.github/workflows/build.yml#L94).

It points to the main branch of the extension-ci-tools repository. Currently, the main branch uses different vcpkg release tags. Any chance to change it to the current release tag instead of main?

hafenkran:
It seems like [Ninja](https://github.com/ninja-build/ninja/issues/1900) causes the problem in combination with very long file paths on Windows. The difference to my repository is the longer directory path based on the `community-extensions` repository name, resulting in warnings like: 

```
CMake Warning in google/cloud/bigquerycontrol/CMakeLists.txt:
  The object file directory

    D:/a/community-extensions/community-extensions/build/release/external/projects/Build/google_cloud_cpp_bigquery/google/cloud/bigquerycontrol/CMakeFiles/google_cloud_cpp_bigquerycontrol_protos.dir/./

  has 197 characters. The maximum allowed full path length for an object file is 250 characters (`CMAKE_OBJECT_PATH_MAX`). Thus, object files such as

    google/cloud/bigquery/v2/biglake_config.grpc.pb.cc.obj

  cannot be reliably created under this directory, potentially breaking the build.
```

~~For now, I've temporarily disabled the Windows build to unblock the build process.~~ However, this might become a recurring issue affecting other projects as well, so there should be a general solution for this, possibly by enabling Windows long-path support. 


**EDIT**: I re-enabled the Windows build again. I did some path changes in my cmake project to reduce the absolute file path length. This workaround lets the build now pass.

hafenkran:
@samansmink @carlopi any1 has the capacity to quickly merge this PR for me? Many thanks in advance!

PR #321, opened 2025-03-18
remove avro (moved to core extensions)

samansmink:
The avro extension will become a core duckdb extension instead, see https://github.com/duckdb/duckdb/pull/16708


carlopi:
I am not completely sure on the path here, this is currently used also for documentation generation / else.

Possibly this can be merged in a preparatory branch for v1.2.2 release?
To be discussed tomorrow.

PR #325, opened 2025-03-19
Update netquack to v1.2.0

hatamiarash7:
## What's Changed
* Add IP Calculator by @hatamiarash7 in https://github.com/hatamiarash7/duckdb-netquack/pull/5. Users can use the `ipcalc` table function to get various details about IP addresses.
* Fix reported problems for extract functions

---

**Full Changelog**: https://github.com/hatamiarash7/duckdb-netquack/compare/1.1.0...1.2.0

### DIFF
--- a/extensions/netquack/description.yml
+++ b/extensions/netquack/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: netquack
   description: DuckDB extension for parsing, extracting, and analyzing domains, URIs, and paths with ease.
-  version: 1.0.0
+  version: 1.1.0
   language: C++
   build: cmake
   license: MIT
@@ -10,7 +10,7 @@ extension:
 
 repo:
   github: hatamiarash7/duckdb-netquack
-  ref: d7f4bece8a3d1e5e74af9965a701332bb3a0800d
+  ref: 665e0fef2550367d44ec9a88ba0343306fc48831
 
 docs:
   extended_description: |
--- a/extensions/netquack/description.yml
+++ b/extensions/netquack/description.yml
@@ -10,7 +10,7 @@ extension:
 
 repo:
   github: hatamiarash7/duckdb-netquack
-  ref: f233da7a72435285d34376fbde31c82b6e5735e4
+  ref: 8094acb2ed43eb20e1ef56f54e91cc49c0788d62
 
 docs:
   extended_description: |
--- a/extensions/netquack/description.yml
+++ b/extensions/netquack/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: netquack
   description: DuckDB extension for parsing, extracting, and analyzing domains, URIs, and paths with ease.
-  version: 1.1.0
+  version: 1.2.0
   language: C++
   build: cmake
   license: MIT
@@ -10,11 +10,11 @@ extension:
 
 repo:
   github: hatamiarash7/duckdb-netquack
-  ref: 665e0fef2550367d44ec9a88ba0343306fc48831
+  ref: f233da7a72435285d34376fbde31c82b6e5735e4
 
 docs:
   extended_description: |
-    This extension designed to simplify working with domains, URIs, and web paths directly within your database queries. Whether you're extracting top-level domains (TLDs), parsing URI components, or analyzing web paths, Netquack provides a suite of intuitive functions to handle all your network tasks efficiently. Built for data engineers, analysts, and developers.
+    This extension designed to simplify working with domains, URIs, IPs, and web paths directly within your database queries. Whether you're extracting top-level domains (TLDs), parsing URI components, or analyzing web paths, Netquack provides a suite of intuitive functions to handle all your network tasks efficiently. Built for data engineers, analysts, and developers.
 
     With Netquack, you can unlock deeper insights from your web-related datasets without the need for external tools or complex workflows.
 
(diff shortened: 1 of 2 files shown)


hatamiarash7:
It's interesting. Several tests have failed on different platforms/architectures. 

The new version has no specific codes that have problems with different architectures or platforms.

I'll check again to find the problem. Would you happen to have any advice?

carlopi:
I think this is due to some platform not running (yet ?!) the unittester.

Error looks to be like:
```
[5/9] (55%): /duckdb_build_dir/test/sql/extract_path.test                       
================================================================================
Wrong result in query! (/duckdb_build_dir/test/sql/ipcalc.test:35)!
================================================================================
SELECT i.IP, ( SELECT hostsPerNet FROM ipcalc(i.IP) ) AS hostsPerNet FROM ips AS i;
================================================================================
Mismatch on row 2, column hostsPerNet(index 2)
NULL <> 1022
================================================================================
Expected result:
================================================================================
127.0.0.1	[25](https://github.com/duckdb/community-extensions/actions/runs/13948183601/job/39042274074?pr=325#step:23:26)4
192.168.1.0/22	1022
================================================================================
Actual result:
================================================================================
1[27](https://github.com/duckdb/community-extensions/actions/runs/13948183601/job/39042274074?pr=325#step:23:28).0.0.1	254
```

where test expect a result different from the generated one.
If you run the test locally (even just as a set of queries), which result do you get?

hatamiarash7:
> I think this is due to some platform not running (yet ?!) the unittester.
> 
> Error looks to be like:
> 
> ```
> [5/9] (55%): /duckdb_build_dir/test/sql/extract_path.test                       
> ================================================================================
> Wrong result in query! (/duckdb_build_dir/test/sql/ipcalc.test:35)!
> ================================================================================
> SELECT i.IP, ( SELECT hostsPerNet FROM ipcalc(i.IP) ) AS hostsPerNet FROM ips AS i;
> ================================================================================
> Mismatch on row 2, column hostsPerNet(index 2)
> NULL <> 1022
> ================================================================================
> Expected result:
> ================================================================================
> 127.0.0.1	[25](https://github.com/duckdb/community-extensions/actions/runs/13948183601/job/39042274074?pr=325#step:23:26)4
> 192.168.1.0/22	1022
> ================================================================================
> Actual result:
> ================================================================================
> 1[27](https://github.com/duckdb/community-extensions/actions/runs/13948183601/job/39042274074?pr=325#step:23:28).0.0.1	254
> ```
> 
> where test expect a result different from the generated one. If you run the test locally (even just as a set of queries), which result do you get?

Tests are completely successful when run locally. This same query returns the correct answer both in test and manual mode. However, in this pipeline, the output value for the second record is `NULL`.

What could have caused the problem?

carlopi:
@hatamiarash7: which duckdb version are you using locally? CI for community-extensions uses, at the moment, v1.2.1

hatamiarash7:
> @hatamiarash7: which duckdb version are you using locally? CI for community-extensions uses, at the moment, v1.2.1

That was `v1.2.0`. I tested `v1.2.1` and [it's the same](https://github.com/hatamiarash7/duckdb-netquack/actions/runs/13950715465/job/39049232467). Everything is working well locally.

There are two test scenarios with the same value ( `192.168.1.0/22` ):

```sql
D SELECT hostsPerNet FROM ipcalc('192.168.1.0/22');
┌─────────────┐
│ hostsPerNet │
│    int64    │
├─────────────┤
│    1022     │
└─────────────┘
 ```
 
 It works.
 
 ```sql
 CREATE OR REPLACE TABLE ips AS SELECT '127.0.0.1' AS ip UNION ALL SELECT '192.168.1.0/22';
 
 SELECT i.IP, ( SELECT hostsPerNet FROM ipcalc(i.IP) ) AS hostsPerNet FROM ips AS I;

┌────────────────┬─────────────┐
│       ip       │ hostsPerNet │
│    varchar     │    int64    │
├────────────────┼─────────────┤
│ 127.0.0.1      │         254 │
│ 192.168.1.0/22 │        NULL │
└────────────────┴─────────────┘
 ```
 
 It does not work. This `ipcalc` function is an **in-out table function**. I think there is a problem with this one in CI.

I have no idea about that ... how can I check this? How can I simulate the CI in a local environment?

carlopi:
Table in-out can be ugly beasts, I am not the right person here, you could consider asking in duckdb discord (https://discord.duckdb.org, there is an #extension channel where others might have seen something similar)

One possibility (not a reccomendation, just pointing this is a way forward) is that this fails CI due to the test failing, but we might also fine with you tweaking the test so it doesn't fail anymore, maybe adding a note that the function is currently experimental (possibly by changing the name or else).

But first, I would ask someone on discord, and / or trying to add a similar test but more punishing (say 1million of rows), hoping this might fail locally. This might be connected to threading, github runners are usually 2 actual threads, so that might trigger different codepaths.

hatamiarash7:
> Table in-out can be ugly beasts, I am not the right person here, you could consider asking in duckdb discord (https://discord.duckdb.org, there is an #extension channel where others might have seen something similar)
> 
> One possibility (not a reccomendation, just pointing this is a way forward) is that this fails CI due to the test failing, but we might also fine with you tweaking the test so it doesn't fail anymore, maybe adding a note that the function is currently experimental (possibly by changing the name or else).
> 
> But first, I would ask someone on discord, and / or trying to add a similar test but more punishing (say 1million of rows), hoping this might fail locally. This might be connected to threading, github runners are usually 2 actual threads, so that might trigger different codepaths.

I tried very hard to reproduce the CI issue but was unsuccessful. For now, I have commented out the relevant test and added a warning that this function is experimental so I can investigate further. I think we can merge this PR now.

PR #327, opened 2025-03-20
Update `read_stat` extension to v0.2.0

mettekou:
@carlopi @samansmink I am introducing support for SAS XPORT and portable SPSS data files with this small update.


carlopi:
Thanks!!

PR #329, opened 2025-03-20
Try to use a different .nullvalue

carlopi:
Maybe @szarnyasg know how to specify an empty nullvalue through strings in bash or in some other way.

Also noted, that `duckdb -help` says:
```
   -nullvalue TEXT      set text string for NULL values. Default ''
```
while current default is `NULL`.

This is likely relevant also in docs generation, whether NULL and empty are not always the same.

### DIFF
--- a/scripts/generate_md.sh
+++ b/scripts/generate_md.sh
@@ -9,7 +9,7 @@ if [ $# -lt 1 ]; then
     exit 1
 fi
 
-DUCKDB_BINARY="$1 -init /dev/null"
+DUCKDB_BINARY="$1 --nullvalue - --init /dev/null"
 
 curl -s https://community-extensions.duckdb.org/downloads-last-week.json -o build/downloads-last-week.json
 


derekmahar:
> Maybe @szarnyasg know how to specify an empty nullvalue through strings in bash or in some other way.

In bash, `duckdb -nullvalue ""` should set the null value to an empty string.

carlopi:
@derekmahar: the problem I have is that I am unsure how to pass empty string in bash scripting, like how to change:
```
export DUCKDB_EXEC="duckdb --nullvalue - -c 'SELECT NULL;' "
$DUCKDB_EXEC
```
to pass proper empty string. This is I guess more a limitation of mine with bash scripting, but if you have a solution it's welcome

derekmahar:
@carlopi, I think we might both agree that quoting and dynamic script evaluation are one of the more confusing and frustrating aspects of shell scripting.

Here are some solutions to your `-nullvalue` empty string problem:

```
$ DUCKDB_EXEC="duckdb -nullvalue '' -c 'SELECT NULL;'"
$ eval $DUCKDB_EXEC
┌───────┐
│ NULL  │
│ int32 │
├───────┤
│       │
└───────┘
$ bash -c "$DUCKDB_EXEC"
┌───────┐
│ NULL  │
│ int32 │
├───────┤
│       │
└───────┘
```

For what it's worth, ChatGPT claims that `bash -c` is safer than `eval` in certain scenarios.

PR #346, opened 2025-03-29
cache httpfs extension release v0.2.1

dentiny:
Hi we could like to release a new version for `cache_httpfs` extension.
I've written a changelog here: https://github.com/dentiny/duck-read-cache-fs/blob/main/CHANGELOG.md

We've tested [unit tests](https://github.com/dentiny/duck-read-cache-fs/tree/main/unit) and [sql test](https://github.com/dentiny/duck-read-cache-fs/tree/main/test/sql) on both macos and linux platform.

### DIFF
--- a/extensions/cache_httpfs/description.yml
+++ b/extensions/cache_httpfs/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: dentiny/duck-read-cache-fs
-  ref: 444c29071816322eb4b1c71c06cb59aa4b93fb85
+  ref: 1797a09c5ea2c55ffbdd0a619ee29e1209a7c36a
 
 docs:
   hello_world: |
--- a/extensions/cache_httpfs/description.yml
+++ b/extensions/cache_httpfs/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: dentiny/duck-read-cache-fs
-  ref: eb2d7f34eac8db50386a764c06b33182b2592b09
+  ref: 444c29071816322eb4b1c71c06cb59aa4b93fb85
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


carlopi:
If I understand the build error about VFS API that happened earlier, the problem is that this extension is developed against `main` duckdb branch instead of vs `v1.2-histrionicus` branch.

Problem is (I think) as follow:
* there are currently 2 (incompatible) duckdb branches, `main` and `v1.2-histrionicus`
* community extensions are currently build only for stable releases, meaning `v1.2.1` and soon the upcoming `v1.2.2`
* depending on API surface an extension touches on, it might be that an extension is at the same time compatible with a wide range of duckdb versions (say `v1.2.1`, `v1.2-histrionicus` and also `main`) or, as in your case, there seems to be changes in FileSystem API that makes so that `main` and `v1.2.1` are different enough not to be compatible with your extension
* I would reccomend, changing (possibly in a branch) the commit of the `duckdb` subfolder to be either `v1.2.1` or the tip of tree of `v1.2-histrionicus` branch, and adapting the code to work against that
* then, send a PR to community-extensions using that hash

Different extensions use different strategies, but what I would recomend is:
* either be anchored to latest released duckdb OR top-of-tree of stable branch

It's this somehow understandable / make some sense?

dentiny:
Thank you @carlopi so much for the quick and detailed reply, I cannot be more grateful for your kindness!

You're completely correct.
I just made a [PR](https://github.com/dentiny/duck-read-cache-fs/pull/175) to temporarily disable features only available at `master` to make it work for v1.2.1
I'm aware we're going to have to [new release](https://duckdb.org/docs/stable/dev/release_calendar.html) pretty soon, so I will likely make a patch for my extension, if the necessary VFS change is picked up.

dentiny:
Thanks @carlopi for the instruction and help along the way, finally get it passed on CI.
I will followup with a documentation somewhere (i.e. extension repository) on the duckdb version requirement and testing instruction, so release process goes much smoother next time.

carlopi:
Great, thanks!

PR #348, opened 2025-04-04
New MSOLAP extension for accessing SSAS and PowerBI

Hugoberry:
An OLEDB for OLAP extension targeting MSOLAP data sources (SSAS and PowerBI). It depends on OLEDB, so it targets only Windows.


Hugoberry:
I have `linux_amd64_gcc4` in the excluded_platforms. Not sure how to proceed with this failed check.

carlopi:
All good, I will have to adapt the check (and docs generation), that currently assumes linux is always a valid platform.
That check is non-critical anyhow, welcome!

PR #350, opened 2025-04-06
Update file_dialog extension to v0.0.2

yutannihilation:
Update the file_dialog extension to v0.0.2. Also, add `function_description.csv`.


carlopi:
Thanks!

yutannihilation:
I found I located `function_description.csv` at a wrong place; it should be under `docs` folder... Sorry, I'll fix in the next release!

carlopi:
That only influences docs generation, so it's also fine to move it whenever.

PR #357, opened 2025-04-08
add parser_tools extension

zacMode:
This adds a new community extension: parser_tools. 

> Exposes functions for parsing referenced tables and usage context from SQL queries using DuckDB's native parser.

**Note on v1.2.2** - _Right before opening this I noticed that 1.2.2 had been released. I don't anticipate any issues updating, but when i tried to follow the [update instructions ](https://github.com/duckdb/extension-template/blob/main/docs/UPDATING.md) i found that [extension-ci-tools](https://github.com/duckdb/extension-ci-tools/branches) doesn't have a 1.2.2. branch yet. Once that exists, i can update the ref to [af7667e140b404c5d02ad70364bd11785aaa6d6f](https://github.com/zacMode/duckdb_extension_parser_tools/commit/af7667e140b404c5d02ad70364bd11785aaa6d6f])._

### DIFF
--- a/extensions/parser_tools/description.yml
+++ b/extensions/parser_tools/description.yml
@@ -10,7 +10,7 @@ extension:
 
 repo:
   github: zacMode/duckdb_extension_parser_tools
-  ref: 560bc45b4bdb43d72dd15cab1073756254e6c055
+  ref: 9818c41f1fb2d7ee65486216a1b4576e5b659df9
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


carlopi:
@zacMode: could you try adding an empty vcpkg.json top level?
Even just:
```
{
}
```
should work around this problem, that is we assume all are vcpkg builds.

Also opening an issue on this limitation would be handy, and we'll sort it out, but easiest way out extension side is just adding a no-op vcpkg.json file.

zacMode:
@carlopi - yes, thanks for the quick assist! Just pushed an update with vpckg.json back in.

carlopi:
Logging off, I will check tomorrow, but thanks for the contribution!

carlopi:
Thanks, and welcome!

PR #366, opened 2025-04-12
wireduck extention

hyehudai:
new wireduck extention - allows dissection of pcap files using tshark behind the scenes.


carlopi:
Thanks, and welcome!

PR #370, opened 2025-04-14
Fix FAISS for 1.2.2

JAicewizard:
Sorry this took so long, had (more of the same) build issues on mingw.

Still working on upstreaming all the patches to vcpkg and faiss to hopefully resolve this more quickly in the future.


carlopi:
Thanks!

PR #374, opened 2025-04-22
Add Ofquack extension: WSDL‑based Oracle Fusion integration

krokozyab:
The Ofquack extension provides seamless integration between DuckDB and Oracle Fusion via WSDL-based SOAP calls. It allows you to run arbitrary SQL queries against Oracle Fusion database directly from DuckDB, inferring column names at runtime and returning all data as VARCHAR columns—as native DuckDB tables and as resultsets that can be directly consumed by downstream applications.


carlopi:
Thanks, and welcome!

This will rebuild, then be published (only DuckDB 1.2.2), then documentation will pick up and show it at https://duckdb.org/community_extensions/extensions/ofquack (404 as of now)

PR #376, opened 2025-04-23
Add a optional relative extension path to description options

joseph-isaacs:


carlopi:
Thanks, this looks to make sense to me, but given this depends on https://github.com/duckdb/extension-ci-tools/pull/172 (interface might change slightly) I think it's better to wait.

joseph-isaacs:
@carlopi Anything we can do to get these two over the line?

0ax1:
@joseph-isaacs We don't rely on this any longer. Shall we just close the PR?

PR #378, opened 2025-04-27
An ODBC extension

Hugoberry:
A cross-platform ODBC extension. 

Heavily dependent on `nanodbc`. 

The existing https://github.com/rupurt/odbc-scanner-duckdb-extension hasn't been updated in the last two years, so this is my take on the subject.

### DIFF
--- a/extensions/nanodbc/description.yml
+++ b/extensions/nanodbc/description.yml
@@ -42,3 +42,5 @@ docs:
     The extension works on Windows, macOS, and Linux platforms and has been tested with SQL Server, MySQL, 
     PostgreSQL, Snowflake, SQLite, and many other databases. All functions use named parameters for better 
     readability and flexibility.
+
+    
--- a/extensions/nanodbc/description.yml
+++ b/extensions/nanodbc/description.yml
@@ -1,7 +1,7 @@
 extension:
-  name: odbc
+  name: nanodbc
   description: Connect to any ODBC-compatible database and query data directly from DuckDB
-  version: 0.1.0
+  version: 0.2.0
   language: C++
   build: cmake
   excluded_platforms: "linux_amd64;linux_arm64;linux_amd64_musl;osx_amd64;wasm_mvp;wasm_eh;wasm_threads;windows_amd64_mingw"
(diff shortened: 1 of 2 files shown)


carlopi:
There seems to be some problem with tests, unsure if additional dependencies are needed or esle.

On the name, we have a very lax policy, but given we have an actual `odbc` driver my personal take is that this might generate some confusion, would it be possible to alter the name?

Thanks!

Hugoberry:
The tests need ODBC drivers, and I can't get away with vcpkg-ing them. I will disable them here with `require-env`.

Regarding the name, I would have gone with `nanodbc`, but I didn't want to infringe on anyone's rights.
Would `odbc_scanner` be ok?

carlopi:
Yes, skipping the tests I think makes sense here. Maybe the error could point to some resource on how to get unstuck (even just the repo readme), but that's minor.

On the name, probably `odbc_scanner` is OK, I will double check on this. One thing, that here might be more relevant than other places, it's very welcome a community provided odbc scanner, but we might eventually want to provide our own, but we will figure then how to resolve naming conflicts.
There is some context here: https://duckdb.org/community_extensions/faq#how-are-naming-collisions-between-extensions-handled.

szarnyasg:
Hi @Hugoberry thanks for this contribution! We discussed it internally and concluded that the name `odbc` or `odbc_scanner` is a bit too generic and might clash with future core DuckDB extensions. Would you mind renaming the extension to e.g. `nanodbc`?

Hugoberry:
I agree with you.
I'm waiting on confirmation from the core developers of `nanodbc` if they are ok with me using the name.

Hugoberry:
I've got the thumbs up from the nanodbc team. 
I've renamed the extension.

Hugoberry:
@carlopi @szarnyasg should I open another PR for this? I can't trigger the checks on this one.

carlopi:
Can you possibly try add a newline or something similar at the yml level? Otherwise I will find out a way myself.

Hugoberry:
I didn't think that was a possibility. Thanks for the recommendation @carlopi

carlopi:
Thanks!

PR #380, opened 2025-04-28
bump bigquery version v0.3.0

hafenkran:
Quick Note: One of the vcpkg dependencies for the linux extensions is sometimes a bit flaky when downloading. It might be the case that you need to restart those steps.


carlopi:
Thanks

PR #385, opened 2025-05-02
Quackformers - better docs and new functionality

martin-conur:
Added Jina Bert embedding implementation. Better docs with examples.


carlopi:
Thanks!

PR #396, opened 2025-05-16
Add DeepSeek 3FS extension

hackmeta:


carlopi:
I think the issue that make CI fail is that it expects a vcpkg.json file (even empty) at the top level in the extension.

This needs fixing on our side, but a workaround to move forward on the extension side would be to add a basic vcpkg.json configuration file (single line with an empty JSON object like: `{ }` should work)

carlopi:
I opened an issue here: https://github.com/duckdb/extension-ci-tools/issues/182

samansmink:

PR #401, opened 2025-05-21
Prepare update to v1.3.0

Maxxen:
This will fail, but should allow us to invoke with the explicit v1.3.0 commit


carlopi:
Indeed this is the expected failure mode.

carlopi (approved):
Thanks, this basically means community extensions enters a state in which CI on PRs will stop working, but thats (temporarily) OK / accepted.

PR #410, opened 2025-05-22
Feature: Support multiple function examples

hatamiarash7:
With the current configuration:

1. We can only have one function example for extensions
2. The output will be an array with one element because of direct export to markdown using the `-markdown` flag and `varchar[]` type

In this PR, the procedure of function extraction is changed to produce a multiline example of functions in the final cell. Also, show all examples as a `<code>` block to look better.

Before:

image

After:

image

Users can add multiple examples for their functions using a `\n` delimiter:

```csv
function,description,comment,example
"extract_domain","Extracting the main domain from a URL","","SELECT extract_domain('a.example.com') as domain;\nSELECT extract_domain('b.example.com') as domain;"
```


carlopi:
Thanks, this makes sense, I would need to test this and come back.

hatamiarash7:
@carlopi, did you have a chance to test it?

I saw that new changes had been applied, and there was a conflict. This was fixed, and I updated again.

samansmink:

PR #411, opened 2025-05-22
Fix: usage example for `fetch_extensions.sh` script

hatamiarash7:
This example is wrong. We should use the `./scripts/fetch_extensions.sh` path.


carlopi:
Thanks for fixing this!!

PR #417, opened 2025-05-24
Serialize examples without escaping apostrophes, etc. in strings

szarnyasg:
Fixes the escaping, see https://github.com/duckdb/duckdb-web/pull/5433/files#r2105758018


carlopi:
I will assume you know what you are up to...

szarnyasg:
yes – it seems to have worked!

PR #419, opened 2025-05-27
Upgrade cache httpfs to v0.3.0

dentiny:
This PR upgrades community extension cache httpfs to v0.3.0
The biggest feature released is filesystem wrapper re-enabled, which allows caching and parallel IO for all duckdb-compatible filesystems.

I have a few other TODO items in mind (will followup when I have more time):
- Check file open info returned from glob, whether we could leverage it as stat cache;
- Check whether we could return file open info for other APIs' return value;
- Check a few github issues on cache effectiveness, toolchain and improvements;
- Check whether to integrate with duckdb native external cache.

### DIFF
--- a/extensions/cache_httpfs/description.yml
+++ b/extensions/cache_httpfs/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: cache_httpfs
   description: Read cached filesystem for httpfs
-  version: 0.2.1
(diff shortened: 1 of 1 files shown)


carlopi:
doc failure is irrelevant (and on us), thanks!

PR #425, opened 2025-06-01
Adding YAML extension

teaguesterling:
This adds an extension that provides a YAML type and parsers and write YAML files similar to the JSON extension.


carlopi:
Thanks!

PR #431, opened 2025-06-03
Update file_dialog

yutannihilation:
Update file_dialog extension to let it be built for DuckDB v1.3.0 (just updated extension-ci-tools submodule).


carlopi:
Thanks!

PR #436, opened 2025-06-06
Cache warming

samansmink:
Add a new mechanism to warm the vcpkg caches


carlopi:
Beautiful script, thanks!

PR #447, opened 2025-06-13
use downgraded aws ci for r2 vcpkg cache

samansmink:


samansmink:
This doesn't run in PR CI, hence blind merge

PR #448, opened 2025-06-15
Add Astro Extension v1.0.0 - Comprehensive astronomical calculations …

bjoernbethge:
…for DuckDB

### DIFF
--- a/extensions/astro/description.yml
+++ b/extensions/astro/description.yml
@@ -1,0 +1,53 @@
+extension:
+  name: astro
+  description: "Comprehensive astronomical calculations and coordinate transformations for DuckDB with Arrow, Spatial, and Catalog integration"
+  version: 1.0.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - bjoernbethge
+
+repo:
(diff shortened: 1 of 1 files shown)


carlopi:
Hi, I am not sure what's up with the CI setup, let me check and come back to you.

extensions/astro/description.yml
```diff
@@ -0,0 +1,53 @@
...
+  version: 1.0.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - bjoernbethge
+
+repo:
+  github: bjoernbethge/astro-duck
+  ref: main
```
carlopi, line 13: `ref:` is intended to be a git sha, pointing to a reference commit, and that then it's bumped on changes

carlopi:
@bjoernbethge: I saw you are iterating on the repo, also great. I can help with the Wasm-side, Windows looks like it's about using a common define.

One possible path is also removing some platforms, like for example https://github.com/duckdb/community-extensions/blob/main/extensions/blockduck/description.yml where windows & wasm platforms are removed.

Then a subsequent PR can re-enable them when they are working.

PR #454, opened 2025-06-16
Add EEA Reference Grid extension

ahuarte47:
This is an extension for DuckDB that adds support for working with the [EEA Reference Grid System](https://sdi.eea.europa.eu/catalogue/srv/api/records/aac8379a-5c4e-445c-b2ef-23a6a2701ef0/attachments/eea_reference_grid_v1.pdf).

The EEA Reference Grid is a standardized spatial grid system used across Europe for environmental data analysis and reporting. It is maintained by the [European Environment Agency](https://www.eea.europa.eu/en) (EEA) and forms the basis for aggregating and exchanging geospatial data in a consistent format.

### DIFF
--- a/extensions/eeagrid/description.yml
+++ b/extensions/eeagrid/description.yml
@@ -1,0 +1,53 @@
+extension:
+  name: eeagrid
+  description: Extension that adds support for working with the EEA Reference Grid System.
+  version: 1.3.1
+  language: C++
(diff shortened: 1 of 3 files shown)


carlopi:
Thanks for the contribution!

ahuarte47:
Thanks @carlopi !!

PR #455, opened 2025-06-17
remove aws cli downgrade

samansmink:


samansmink:
ci is not run for this change, hence insta-merge

PR #461, opened 2025-06-20
Update file_dialog and st_read_multi

yutannihilation:
Update file_dialog extension and st_read_multi extension.

While duckdb-rs v1.3.1 is not out yet, I believe my extensions doesn't require the updated source because there are no API changes related to the extension API. I quickly confirmed this can be loaded on DuckDB v1.3.1 on my local.


yutannihilation:
> `ValueError: cannot have multiple descriptors changed or packages with spaces in their names`

Ah, I didn't know this, sorry. I'll retry later...

PR #462, opened 2025-06-20
Update file_dialog to v1.3.1

yutannihilation:


yutannihilation:
I confirmed the extension can now be `INSTALL`ed and `LOAD`ed, but it's not listed here yet. Are there anything that I still have to fix...?

https://duckdb.org/community_extensions/list_of_extensions

PR #466, opened 2025-06-23
removed avro

c-herrewijn:
Removed avro from community extensions, since it is a regular extension now: https://github.com/duckdb/duckdb-avro


carlopi:
Thanks!

PR #480, opened 2025-07-08
add lua extension

isaacbrodsky:
Not sure this makes sense to merge until DuckDB 1.4.0 is released, as the extension is only compatible with that version. But I guess that's what ref_next is for?


isaacbrodsky:
This should now be ready for review

carlopi:
Thanks!

PR #487, opened 2025-07-10
Adding an extension to work with markdown files and parse markdown sections

teaguesterling:
See: https://github.com/teaguesterling/duckdb_markdown


carlopi:
I think we have a requirement that each PR should touch at most a given descriptor, could you adapt and force push?

teaguesterling:
Ah! Sorry about that! Accidentally created it off the changes with the YAML extension. Will do!

carlopi:
@teaguesterling: I think I know where the problem is: `cmark-gfm` vcpkg package is not available in the default vcpkg commit we use, that is `ce613c4137` corresponding to tag `2025.04.09`.

You can provide in the descriptor a specific `vcpkg_commit`, see https://github.com/duckdb/community-extensions/blob/c2886a2ff831fc42bbedef1a6afbd3ede8cea39e/extensions/quack/description.yml#L14, that should move this further.

teaguesterling:
Thank you @carlopi for taking the time to troubleshoot that for me! I had that fixed in the extension's workflow, but didn't realize I needed to put it here as well. Much appreciated!

carlopi:
Thanks!

PR #491, opened 2025-07-15
Update chsql for v1.3.2

lmangani:


carlopi:
@lmangani: I think fix at https://github.com/duckdb/extension-ci-tools/pull/237 should have this now working, I restarted CI, I am not completely sure if the right version of extension-ci-tools will be picked up, if not (= same failure), I will check again later, or please ping.

carlopi:
It took me a bit to see it, but I have the suspect the problem comes from your extension Makefile, that looks like:
```
PROJ_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))chsql/
```
instead of
```
PROJ_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
and it's somehow assumed that `$(PROJ_DIR)/extension-ci-tools` is a valid folder.

Could you change PROJ_DIR, fix the build, and re-submit? Thanks a ton

lmangani:
> It took me a bit to see it, but I have the suspect the problem comes from your extension Makefile, that looks like:
> 
> ```
> PROJ_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))chsql/
> ```

True, but our project code is under chsql/ and not in the root like other extensions so changing this would fail the build. This was never an issue until now and this hasn't been changed recently.

carlopi:
If you look at the line here: https://github.com/duckdb/community-extensions/actions/runs/16351243587/job/46247100914?pr=491#step:23:46, it says it's looking for
```
DVCPKG_OVERLAY_TRIPLETS=D:/a/community-extensions/community-extensions/chsql/extension-ci-tools/toolchains
```
that is not there (given it's at community-extensions/community-extensions/extension-ci-tools)

But I see the point, it's not on you to change `PROJ_DIR`, but in us using a different variable where we do:
```
TOOLCHAIN_FLAGS:=${TOOLCHAIN_FLAGS} -DVCPKG_OVERLAY_TRIPLETS=${PROJ_DIR}extension-ci-tools/toolchains
```
https://github.com/duckdb/extension-ci-tools/blob/1697246177f0cf098ed54838aa77c2deedd43a8f/makefiles/duckdb_extension.Makefile#L93

I will have a look at fixing that.

lmangani:
Thanks Carlo! Happy to make changes if this turns out to be too complex. Just not understanding how it only confuses this one builder and not the others

carlopi:
I think https://github.com/duckdb/extension-ci-tools/pull/240 might be the solution, need some time to double check / go through CI. Thanks for your comment.

samansmink:
I think the way the chsql extension is setup directory-structure-wise is a little non-standard and currently not supported. See my comment here https://github.com/duckdb/extension-ci-tools/pull/240#pullrequestreview-3038184366.

@lmangani would you mind switching the chsql extensions to use the standard directory structure from the extension-template? 

Alternatively if you feel you really need to use an alternate directory structure I'm happy to guide you with a PR, however we don't really have the cycles available to work on this ourselves right now.

lmangani:
> @lmangani would you mind switching the chsql extensions to use the standard directory structure from the extension-template?

Sure we could, although I still don't understand why this would only impact one builder. I'll just exclude windows builds for now since we have users out there and this surprise factor is slightly inconvenient and difficult to improvise.


> I'm happy to guide you with a PR, however we don't really have the cycles available to work on this ourselves right now.

Same here unfortunately this type of surprise is affecting both sides. We had zero warnings and trying to figure this out.

samansmink:
> We had zero warnings and trying to figure this out.

I feel you, this is something we are working on improving for next releases

lmangani:
Thanks! If the current build is acceptable to release, we'll target normalizing the repo and restoring windows next round 👍

samansmink:
sounds good!

PR #494, opened 2025-07-18
Update splink_udfs to 0.0.2

RobinL:
I have added two more functions to this extension.

I note that the builds all passed on https://github.com/moj-analytical-services/splink_udfs/pull/4 see [here](https://github.com/moj-analytical-services/splink_udfs/actions/runs/16368072181)

but that the 'Tidy check' failed.  I've done a bit of research that suggests this isn't a problem, but I'm not very experienced with c++, so I'm not sure.

If this is a problem I'd appreciate any tips/pointers on what I'm doing wrong.  I had quite a few issues getting `CMakeLists.txt` correct with the new `utf8proc` dependency, so this is the most likely cause of the issue.  

Everything works locally on my macbook (M4).

Thank you!

### DIFF
--- a/extensions/splink_udfs/description.yml
+++ b/extensions/splink_udfs/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: splink_udfs
-  description: Record linkage functions for use in Splink
-  version: 0.0.1
+  description: Phonetic and text normalization functions for record linkage.
+  version: 0.0.2
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
I think the issue with clang-tidy is the same as https://github.com/duckdb/extension-ci-tools/issues/223, you can subscribe there, unsure of timing for solving it.

You should be able to disable via passing explicit list of checks, like changing this line https://github.com/moj-analytical-services/splink_udfs/blob/main/.github/workflows/MainDistributionPipeline.yml#L39 to `format_checks: 'format'`, or you can also just skip the block completely, it's not compulsory

RobinL:
Thank you!

PR #498, opened 2025-07-19
feat: add airport extension

rustyconover:
Add the airport extension.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,7 @@ jobs:
       COMMUNITY_EXTENSION_DEPLOY: ${{ steps.parse.outputs.COMMUNITY_EXTENSION_DEPLOY }}
       COMMUNITY_EXTENSION_EXCLUDE_PLATFORMS: ${{ inputs.more_excluded }}${{ steps.parse.outputs.COMMUNITY_EXTENSION_EXCLUDE_PLATFORMS }}
       COMMUNITY_EXTENSION_REQUIRES_TOOLCHAINS: ${{ steps.parse.outputs.COMMUNITY_EXTENSION_REQUIRES_TOOLCHAINS }}
+      COMMUNITY_EXTENSION_TEST_CONFIG: ${{ steps.parse.outputs.COMMUNITY_EXTENSION_TEST_CONFIG }}
       COMMUNITY_EXTENSION_VCPKG_URL: ${{ steps.parse.outputs.COMMUNITY_EXTENSION_VCPKG_URL == '' && 'https://github.com/microsoft/vcpkg.git' || steps.parse.outputs.COMMUNITY_EXTENSION_VCPKG_URL }}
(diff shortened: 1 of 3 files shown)


rustyconover:
Today, I unleash it. Powered entirely by DuckDB, Arrow, duct tape, and blind optimism. 🦆✨

Let the bug reports waddle in! 🐛🦆

carlopi:
I am not really sure on the segfault after the test report in https://github.com/duckdb/community-extensions/actions/runs/16392157077/job/46319554618?pr=498#step:24:97, I suspect either it's on us, where the unittester itself has some problem dealing with some corner-case, or it might be connected to some destructor called on the extension side (I don't see how though).

I will have a look tomorrow.

Did you see something like that before?
Also considering https://github.com/duckdb/community-extensions/issues/500, that no meaningful test is currently run, maybe it's also OK to merge in any case, but then you would need to skip testing on this platform on the Makefile side?

rustyconover:
> I am not really sure on the segfault after the test report in https://github.com/duckdb/community-extensions/actions/runs/16392157077/job/46319554618?pr=498#step:24:97, I suspect either it's on us, where the unittester itself has some problem dealing with some corner-case, or it might be connected to some destructor called on the extension side (I don't see how though).

I think it is possible that `unittester` has a bug where if all tests are skipped a segfault occurs.  There isn't any destructor code in Airport explicitly.

I'm going to add a single SQL test that isn't subject to the presence of an environment variable to work around it the possible bug.

In the Airport GitHub repo, tests pass fine.  I worked a long time on creating them and making them pass. I'd love to have them run in production.

> I will have a look tomorrow.
> 
> Did you see something like that before? Also considering #500, that no meaningful test is currently run, maybe it's also OK to merge in any case, but then you would need to skip testing on this platform on the Makefile side?

I'd really like to have the tests run for Airport just like any other extension, I think there may be some GitHub Actions changes that need to be made so that the right data get passed down to the action invocation.  In the Airport repo everything works great with the `extension-ci-tools` call to the action, I'm not sure why it's not working here in this repo.  If you could look at it I would appreciate.  If not I'll try to look at it in a few days.

Thanks as always for your help @carlopi,

Rusty

carlopi:
Thanks for the iteration, I do think there might be a bug on our side, we'll investigate.

If you want to work around this on your side you can provide an extra empty target in Makefile, overriding the failing `test_release_internal`, basically something like:

```
test_release_internal:

```

This is only work-around the current yet to be tracked problem AND the fact that ENV-parameter passing is not yet there.

carlopi:
@rustyconover: I still have a hard time deciding which side it is, and maybe it's faster if you manage to workaround this extension side somehow.
With a simpler reproducer would be handy, but as things stand now it's quite possible it's something on the extension side that I can't see.

rustyconover:
Hi @carlopi,

Thanks for looking at it today.

Can you help me just get the environment variable propagation working?  If so that will match the development repo and everything should work great.

Rusty

carlopi:
Thanks!

PR #515, opened 2025-07-27
Upgrade cache httpfs to 0.5.0

dentiny:
This PR upgrades cache httpfs extension to v0.5.0, a few changes listed:
- Upgrade dependencies: duckdb-httpfs and extension-ci
- Increase default IO request size from 64KiB to 512KiB
  + This is half a bug, I always know 2MiB request size is the sweetspot for object storage, but forgot to update value before release
- Allow multiple on-disk cache directories
- Attempt to cache file metadata from `ExtendedFileInfo`


carlopi:
Thanks

PR #523, opened 2025-08-07
Upgrade cache httpfs to 0.6.0

dentiny:
Upgrade cache httpfs to v0.6.0, changelog see https://github.com/dentiny/duck-read-cache-fs/blob/main/CHANGELOG.md#060

The biggest change is to provide observability for cache miss count by in-use exclusive resource.


dentiny:
> 19.03 fatal: cannot create directory at 'ports/intel-mkl': No space left on device

Looks like CI issue, should be retried :)

PR #530, opened 2025-08-12
Migrate httpserver to query.farm

lmangani:
First of many! PR lurkers hear it first - we're migrating all quackscience > query.farm and joining forces on our community extensions with @rustyconover


samansmink:
@lmangani is this good to go?

PR #552, opened 2025-09-08
Add snowflake extension

venkata-chikkam:
- Adds the DuckDB Snowflake extension to the community extensions repository.

- The extension integrates DuckDB with Snowflake using Arrow ADBC drivers, providing both table functions (snowflake_scan) and storage extension capabilities (ATTACH).

- Extension details and usage examples are included in the description.yml file.

### DIFF
--- a/extensions/snowflake/description.yml
+++ b/extensions/snowflake/description.yml
@@ -10,7 +10,8 @@ extension:
 
 repo:
   github: iqea-ai/duckdb-snowflake
-  ref: 6b80fdc34cd3715a8c6167dfac2f662001351e60
+  ref: 82a1fac62130046eda51dc75e9e9abdf89e956f5
+  ref_next: eaa953343f5cc0d1e53704b751fea5e70f556bf4
 
 docs:
   hello_world: |


samansmink:
Hey @venkata-chikkam thanks for the PR, this looks really exciting!

Your ref and ref_next are pointing to the same commit? they both appear to use the v1.4.0 entrypoint, so this can not pass CI right now (but it may be able to in a few days when duckdb v1.4.0 is out)

venkata-chikkam:
Hi @samansmink, great to hear from you and thanks for the feedback!

I updated the descriptor so ref and ref_next now point to separate commits of the extension, with each branch building against its respective DuckDB version (`ref` for `1.3.2`, `ref_next` for `1.4`), which should resolve the CI issue.

I also wanted to ask about dependencies: since the extension requires `libadbc_driver_snowflake.so` at runtime (otherwise it fails with `Failed to initialize database. Unknown ADBC error`), would you recommend bundling the .so in the artifact zip or leaving it for the end user to download separately?

Thanks,
Venkata

samansmink:
All green, go go go! 🟢 🟢 🟢 

>  would you recommend bundling the .so in the artifact

linking extensions with shared libraries is something we have tried to avoid as much as possible, so we don't have much experience with that. For best user experience it would be nice if the library could be bundled or loaded automatically. If that's tricky you could add some installation instructions to your community extension docs

PR #554, opened 2025-09-10
Update to splink_udfs 0.0.9, add 1.4

RobinL:
Updated and simplified address cleaning functions - thanks very much! 

Now works with 1.4


carlopi:
Thanks!

PR #564, opened 2025-09-16
Dont use extension ci tools for matrix

samansmink:


carlopi:
Only comment: now it implies there are no `deploy_matrix.json` top level in extension-ci-tools, but that's fine (could be slightly cleaner to keep in a separate folder)

But this is simple / good enough

PR #573, opened 2025-09-16
fix: lindel 1.4

rustyconover:


extensions/lindel/description.yml
```diff
@@ -10,8 +10,7 @@ extension:
     - rustyconover
   name: lindel
   requires_toolchains: rust
-  version: 1.0.1
```
carlopi, line 13: I think the version should be kept

rustyconover: ok

rustyconover: I'll get things updated.

carlopi:
Thanks!

PR #576, opened 2025-09-16
fix: rapidfuzz 1.4

rustyconover:


carlopi:
Apparently there is no `36ab36362092c05942e9128f898eba550d322bc3` sha in the relevant repo, could you double check?

carlopi:
Thanks!

PR #579, opened 2025-09-16
Swap ref_next to ref, and bump to duckdb 1.4.0

carlopi:
Reading community-extensions for the v1.4.0 cycle!


carlopi:
(this is expected to fail on startup due to the multiple extensions that are changed)

PR #586, opened 2025-09-17
Update nanoarrow for v1.4.0

evertlammerts:


carlopi:
Thanks!

carlopi:
I'll merge, but we are in the process of performing some maintenance, I might have to restart later.

PR #587, opened 2025-09-17
Add endpoint config

hannes:


hannes:
I've added the secrets to the repo config

hannes:
lets go

PR #589, opened 2025-09-17
Bump quack

hannes:


hannes:
CC @carlopi, seems the action was not triggered

carlopi:
Thanks, looks good, let's check the deploy step

PR #593, opened 2025-09-17
Create observefs extension

dentiny:
This PR creates duckdb observefs extension.

Motivation:
As a developer, I feel painful when debugging duckdb IO perf issues. For example, it's hard to tell whether a slow query comes from query execution itself, or it's from slow IO.
`EXPLAIN ANALYZIZ` only tells the number of IOs, or verbose logging is also not helpful here.
The issue becomes larger when I'm accessing different objects scattered around different buckets at different regions.

So in initial extension release, 
- I add very basic latency stats to open and read operations (the two I care about most)
- Bucket-level stats are accessible -- overall it doesn't make sense to mix all stats for different buckets together
- IO latency outliers are recorded separately
- I tried my best to keep the good parts for cache httpfs: compatible with https

The extension is written in half a day, so still in very basic stage. A few TODO items:
- Better quantile stats; current implementation is built upon histogram which is very coarse especially when data points small
- More IO operations supports, list will be my next
- Error stats report, also aggregate by `errno` if possible
- No support for v1.4 for now, since it's introducing breaking API changes, I will take a look later

### DIFF
--- a/extensions/observefs/description.yml
+++ b/extensions/observefs/description.yml
@@ -1,0 +1,24 @@
+extension:
+  name: observefs
+  description: Provides IO observability to filesystem
+  version: 0.0.1
+  language: C++
+  build: cmake
+  license: MIT
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64;windows_amd64_mingw"
+  maintainers:
+    - dentiny
+
+repo:
+  github: dentiny/duckdb-filesystem-observability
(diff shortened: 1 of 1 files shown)


dentiny:
```sh
/duckdb_build_dir/src/include/observefs_extension.hpp:9:14: error: 'void duckdb::ObservefsExtension::Load(duckdb::DuckDB&)' marked 'override', but does not override
    9 |         void Load(DuckDB &db) override;
      |              ^~~~
```
Emmm looks weird to me

carlopi:
Thanks!

PR #607, opened 2025-09-21
Upgrade cache httpfs

dentiny:
This PR upgrades cache-httpfs extension, which includes a few updates:
- Upgrade duckdb and httpfs to v1.4 and its compatible version
- Add LRU-based on-disk cache block eviction


carlopi:
Thanks!

PR #608, opened 2025-09-21
Upgrade observefs to v0.2

dentiny:
This PR upgrades observefs, which includes a few updates:
- Support filesystem wrapping, for example, wrap azure filesystem from other extensions
- Implement quantile estimation, and treat differently for small data points and large ones
- Refine display output if there's no interested IO operations happening in the current filesystem


carlopi:
Thanks!

PR #610, opened 2025-09-22
update zipfs for globbing bug

isaacbrodsky:
Fixes isaacbrodsky/duckdb-zipfs#45 via isaacbrodsky/duckdb-zipfs#46


carlopi:
Thanks!

PR #620, opened 2025-09-25
bump vortex to 0.53.0 & duckdb v1.4.0

0ax1:


0ax1:
Merged in latest main, including https://github.com/vortex-data/community-extensions/commit/cc03b3a17eb15eeaa2ee9486e1a6ac926fa83350 to the fix the deploy step. @carlopi

carlopi:
I think the deploy should be now fixed also via https://github.com/duckdb/community-extensions/pull/621, and this PR I would assume should go through, but we will see.

Older PRs would need to be manually re-triggered, we'll do so later.

carlopi:
Thanks!

PR #632, opened 2025-09-30
Add cache_httpfs extension

dentiny:
I found a lot of people in duckdb community is asking for connection pool feature, I also met the same issue in my work, so I wrote this extension based upon httpfs to resolve the problem.

Some high-level ideas:
- 100% compatible with httpfs, so there's no migration pain point for users
- I rewrite (or say, plan to rewrite) the whole duckdb http layer and build everything upon libcurl and a polling engine; right now I'm using epoll and leverage reactor model;
- A few key features have been implemented, including http/2, TCP connection pool, async IO (future/promise style), etc.

There're several TODOs which I will very likely finish within the week:
- As of now, only HEAD and GET HTTP requests have been integrated into multi-curl, which I think are the two most important ones; I will do the later this week.
- Resolve known perf issue: https://github.com/dentiny/duckdb-curl-filesystem/issues/14

### DIFF
--- a/extensions/curl_httpfs/description.yml
+++ b/extensions/curl_httpfs/description.yml
@@ -5,7 +5,7 @@ extension:
   language: C++
   build: cmake
   license: MIT
-  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64;windows_amd64_mingw"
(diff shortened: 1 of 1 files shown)


carlopi:
Could you have a look at the MacOs failure? Thanks!

dentiny:
> Could you have a look at the MacOs failure? Thanks!

Thanks for the quick reply! 
(not a mac developer) I just realize `epoll` is a linux-specific syscall which is not available on macos; so simply disable it for now, will checkout `kqueue`.

carlopi:
Thanks!

PR #633, opened 2025-09-30
Bug fix patch for cache httpfs v0.7.1

dentiny:
This PR is a bug fix patch:
- Fix a segfault bug
  + reference: https://github.com/dentiny/duck-read-cache-fs/pull/250
- Disable extended filesystem interfaces which we cannot access in the extension, so no open or list slip through cached filesystem API calls


carlopi:
Thanks!

PR #644, opened 2025-10-03
curl httpfs add macos support

dentiny:
This PR adds macos support for curl httpfs, essentially use `kqueue` as polling engine.


carlopi:
Thanks!

PR #649, opened 2025-10-06
Upgrade observefs to 0.3

dentiny:
This upgrade 
- Add metrics on read request size
- Fix quantile calculation on small number of data points
- Fix http request metrics reporting

Reference: https://github.com/dentiny/duckdb-filesystem-observability/blob/main/CHANGELOG.md#030


carlopi:
Thanks!

PR #652, opened 2025-10-06
Add dns extension

tobilg:
I created a new extension for (reverse) dns lookups from within DuckDB


carlopi:
Thanks! And welcome!

This we be now built, to be available in the next hour or so.

PR #654, opened 2025-10-07
Patch observefs to v0.3.1

dentiny:
This PR patches observefs to v0.3.1, two main changes:
- Fix a bug that all httpfs filesystems are registered twice! One as raw filesystem, one wrapped in observefs
- Add a list filesystem function, so users could know what's available filesystem before they wrap


carlopi:
Thanks!

PR #660, opened 2025-10-07
Remove double install instructions

Ian-Fogelman:
Hi DuckDB, please merge this PR to avoid confusion in the psyduck docs. I didn't realize that install / load would be taken care of in the `Installing and Loading` section of the page: 

https://duckdb.org/community_extensions/extensions/psyduck.html

image


carlopi:
Thanks!

PR #663, opened 2025-10-07
Upgrade curl_httpfs to v0.2.1

dentiny:
No functionality change, just upgrade duckdb to v1.4.1


carlopi:
Thanks!

For the future: bumping the submodule should be done automatically by CI, so if you have not changed any code it would not be strictly needed.

dentiny:
> For the future: bumping the submodule should be done automatically by CI, so if you have not changed any code it would not be strictly needed.

That would be nice!

PR #669, opened 2025-10-08
Bump boilstream to v0.2.0

dforsber:
boilstream v0.2.0 - A Secure Remote DuckDB Secrets Store

- PKCE one-time short term bootstrap token exchange to longer living in-memory only session token
- Integration and unit tests


carlopi:
Thanks!

PR #722, opened 2025-10-14
Upgrade observefs to v0.3.3

dentiny:
Two changes:
- (major) Fix issues where `cache_httpfs` cannot be used together with `observefs`
  + cache httpfs upgrade will follow through later this week, have some other updates I would like to check in together
- (minor) Update filesystem list function from scalar function to table function for better display without `unnest`


carlopi:
Thanks!

PR #723, opened 2025-10-14
Rework Vortex description

szarnyasg:


szarnyasg:
Thx @carlopi for the review!

carlopi:
@0ax1: unsure what do you think here, it's docs only

0ax1:
> @0ax1: unsure what do you think here, it's docs only

Sure, looks good!

PR #731, opened 2025-10-15
Bump boilstream to 0.3.1 - session resumption support

dforsber:
boilstream v0.3.1 - A Secure Remote DuckDB Secrets Store

- Added support for session resumption with on-disk refresh_token
- NOTE: Server informs about session resumption capability on http header


carlopi:
Thanks!

PR #740, opened 2025-10-17
Add encoding extension v0.1.0

onnimonni:
Hey,

I had following issue on trying to parse non UTF-8 encoded json:

```sql
SELECT json(decode('\x5B\x7B\x22Langue\x22\x3A\x5B\x22Fran\xE7ais\x22,\x22English\x22\x5D\x7D\x5D')) as json;

Conversion Error:
Failure in decode: could not convert blob to UTF8 string, the blob contained invalid UTF8 characters
```

This extension tries to alleviate this problem by providing `force_utf8_encoding` function which has access to all different encodings from the rust package `encoding_rs`: https://docs.rs/encoding_rs/latest/encoding_rs/.

When this extension is loaded this works easily:
```sql
SELECT json(force_utf8_encoding('\x5B\x7B\x22Langue\x22\x3A\x5B\x22Fran\xE7ais\x22,\x22English\x22\x5D\x7D\x5D')) as json;
┌─────────────────────────────────────┐
│                json                 │
│                json                 │
├─────────────────────────────────────┤
│ [{"Langue":["Français","English"]}] │
└─────────────────────────────────────┘
```

Hopefully this will help others who would need to mangle data in weird encodings like I needed.

Thanks a lot for @tobilg and his `dns` extension. It was very easy for claude to create this extension by using the dns extension as a starting point 🙇.

### DIFF
--- a/extensions/encoding/description.yml
+++ b/extensions/encoding/description.yml
@@ -1,0 +1,189 @@
+extension:
+  name: encoding
+  description: The Encoding Extension enables text encoding detection and UTF-8 conversion from within DuckDB
+  version: 0.1.0
+  language: Rust
+  build: cargo
+  license: MIT
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;linux_amd64_musl;"
+  requires_toolchains: "rust;python3"
+  maintainers:
+    - onnimonni
+
+repo:
+  github: onnimonni/duckdb-encoding
+  ref: 8d211ed71b3a212468ff46e83dbec847e56f2467
+
+docs:
+  hello_world: |
+    -- Forces text with potentially mixed or unknown encoding to UTF-8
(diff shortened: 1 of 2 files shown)


carlopi:
Hi! Great on adding, the more the merrier, only question is on name, given there is an `encodings` core extension already (likely with a different feature set), and I can see confusion arising

carlopi:
Letting CI having a go at this, need to figure out if name is fine or too similar.

carlopi:
I'll be merging, if reasons to move emerge we can think more about that later.

Thanks, and welcome!

onnimonni:
I'm happy to rename this if needed. Also @Maxxen mentioned in the slack that `encodings` extension will eventually expose similiar functionality in the future: https://discord.com/channels/909674491309850675/921073327009853451/1428648055648944189

When it's available there I think we can mark this extension as deprecated and point to those functions instead 👍

PR #745, opened 2025-10-18
Upgrade cache httpfs v0.9.0

dentiny:
This PR upgrades cache httpfs to v0.9.0, which includes two additional features:
- Add an in-memory cache layer to on-disk cache reader, so applications and people don't need to access disk in most cases
- Add a few cache config read table functions, to group related configs together.

### DIFF
--- a/extensions/cache_httpfs/description.yml
+++ b/extensions/cache_httpfs/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: dentiny/duck-read-cache-fs
-  ref: 699aab185e0ad151a2589fb46891701ff9499232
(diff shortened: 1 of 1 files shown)


dentiny:
Hmm, it cannot be reproduced locally.
```sh
[~/duck-read-cache-fs] (main) 
vscode@9a5948bd4884$ make test_release
./build/release/"/test/unittest" "test/*"
Filters: test/*
[17/17] (100%): test/sql/insufficient_disk_space.test                           
===============================================================================
All tests passed (6579 assertions in 17 test cases)
```

dentiny:
Should be good to go, though I still don't understand why I cannot reproduce CI failure on my local testing. :(

carlopi:
Thanks!

PR #748, opened 2025-10-20
Upgrade observefs to v0.3.0

dentiny:
This PR upgrades observefs to v0.3.0, which includes a few fixes and additions:
- Fix read operations latency record, previous only open operation gets recorded
- Add additional latency collection for glob and get file size, so all read operations get collected
- Add a rough cache access record for duckdb external file cache, including cache hit/miss/partial hit
- Remove TCP connection status query function, which has been removed to another extension `curl_httpfs`


carlopi:
Thanks!

PR #749, opened 2025-10-20
Updates for the dns extension v0.4.0

tobilg:
Added a new `set_dns_cache_size(bigint)` function that allows to set the dns query result cache size,


carlopi:
Thanks!

PR #753, opened 2025-10-21
fix: bump airport

rustyconover:


rustyconover:
@carlopi could you please re-run this it seems like a transient error on windows, otherwise everything is fine.

carlopi:
Thanks!

PR #754, opened 2025-10-22
feat: bump vortex to 0.54.0

0ax1:


0ax1:
@carlopi I think the PR should be good to go.

carlopi:
Indeed, thanks!

PR #761, opened 2025-10-25
[mlpack] Expanded description.yml with docs, use updated repo

eddelbuettel:
This PR expands the yaml file by adding a `docs` section which should make the now-generated web page less bare :laughing: 

It also rolls forward to a newer commit sha1 as another predict function has been added.


carlopi:
Thanks!

PR #764, opened 2025-10-27
[mlpack] Update

eddelbuettel:
This adds
- macOS builds
- more unit tests
- extension to model table so e.g. linear regression can store coefficients
- edits to README

PR #765, opened 2025-10-27
[mlpack] Turn on macOS builds

eddelbuettel:
This was of course supposed to be part of #764.  "Better late than never."   Sorry for making you approve twice.


carlopi:
Thanks!

PR #769, opened 2025-10-28
[mlpack] Update documentation, mention macos binary

eddelbuettel:
This updates the documentation to mention macOS binaries which the original (and follow-up) commits failed to do.  So this round really is one of three.


carlopi:
Thanks!

PR #778, opened 2025-11-03
Improve the extension description

habedi:
Hi,

I'm sending this PR to update the `extension/infera/description.yml` and also include function metadata.


carlopi:
Thanks!

PR #779, opened 2025-11-03
Update snowflake extension ref_next to main branch for DuckDB v1.4.1

venkata-chikkam:
Updates the snowflake extension to use the new main branch commit (fba82f9) for `ref_next`, which builds against DuckDB v1.4.1.

This PR:

1. Keeps `ref` pointing to stable branch (abf57bc) for DuckDB v1.4.0 users
2. Updates `ref_next` to main branch (fba82f9) for DuckDB v1.4.1 users
3. Will trigger a rebuild of v1.4.0 binaries with the latest features

Both commits include:

- snowflake_query() function
- Multiple authentication methods (Password, SSO, Key Pair)
- Query pushdown optimization
- All latest features

This ensures users on both DuckDB v1.4.0 and v1.4.1 get the updated extension with snowflake_query instead of the old snowflake_scan function.

### DIFF
--- a/extensions/snowflake/description.yml
+++ b/extensions/snowflake/description.yml
@@ -10,7 +10,7 @@ extension:
 
 repo:
   github: iqea-ai/duckdb-snowflake
-  ref: abf57bcc2512dae834257310af4dedddd72f6ce2
+  ref: fba82f97ca11cb279603c2ccff353b3db4fe1e32
   ref_next: fba82f97ca11cb279603c2ccff353b3db4fe1e32
 
 install_notes: |


carlopi:
Hi, I am fine with merging the PR, but the end result will be that only the sha referenced as `ref` will be used.

We currently build community extensions only for the latest supported duckdb version, that at the moment is 1.4.1, and users of duckdb v1.4.0 are encouraged to move forward.

Let me know if you want this merged, or you want (as I would probably suggest), unify `ref` and `ref_next`.

venkata-chikkam:
Thanks for the clarification! I’ve updated the descriptor so both ref and ref_next point to fba82f97ca11cb279603c2ccff353b3db4fe1e32, so the Snowflake extension will build from that commit for the latest DuckDB (v1.4.1). Feel free to merge when ready, and let me know if a different commit is preferred.

carlopi:
Thanks!

PR #784, opened 2025-11-05
Add duckherder extension

dentiny:
This PR adds initial version duckdb community extension `duckherder`, which runs query on remote server.
- `duckherder` storage extension is implemented on client side to communicate with server and observability
- `duckling` storage extension is used on server side to inject and execute queries
- though unimplemented for now, I plan to investigated distributed query via `duckling`
- query results are passed back to client side via arrow flight

### DIFF
--- a/extensions/duckherder/description.yml
+++ b/extensions/duckherder/description.yml
@@ -5,7 +5,7 @@ extension:
   language: C++
   build: cmake
   license: MIT
-  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64;windows_amd64_mingw"
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64;windows_amd64_mingw;osx_amd64;osx_arm64"
   requires_toolchains: parser_tools
   maintainers:
     - dentiny


dentiny:
```sh
Building thrift:x64-linux-release@0.20.0#1...
/vcpkg/triplets/community/x64-linux-release.cmake: info: loaded community triplet from here. Community triplets are not built in the curated registry and are thus less likely to succeed.
CMake Error at scripts/cmake/vcpkg_find_acquire_program.cmake:166 (message):
  Could not find flex.  Please install it via your package manager:

      sudo apt-get install flex
Call Stack (most recent call first):
  ports/thrift/portfile.cmake:9 (vcpkg_find_acquire_program)
  scripts/ports.cmake:206 (include)
```
I'm not sure what should I do here...

carlopi:
I think the path is adding `parser_tools` as build time dependency. See for example how that's added here: https://github.com/duckdb/community-extensions/blob/de74110c0003573036bce36863f839785aaac02a/extensions/airport/description.yml#L13

dentiny:
> I think the path is adding `parser_tools` as build time dependency. See for example how that's added here:
> 
> https://github.com/duckdb/community-extensions/blob/de74110c0003573036bce36863f839785aaac02a/extensions/airport/description.yml#L13

Thank you for the quick reply and help! Let me have a try.

dentiny:
Linux build passes, still unsure on the macos issue; cannot repro on my local (though not the same macos SDK version).
Plan to check later, temporarily disable for now.

carlopi:
Thanks!

PR #805, opened 2025-11-13
Upgrade cache httpfs to 0.9.3

dentiny:
This PR upgrades duckdb, httpfs and extension-ci-tools to v1.4.2.
The tricky thing is, main branch for httpfs extension is not compatible with duckdb v1.4.2, so I stole it from Carlo's ongoing PR. :)

### DIFF
--- a/extensions/cache_httpfs/description.yml
+++ b/extensions/cache_httpfs/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: cache_httpfs
   description: Read cached filesystem for httpfs
-  version: 0.9.2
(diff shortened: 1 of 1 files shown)


carlopi:
I think `v1.4-andium` branch + patches has to be compatible with duckdb `v1.4-andium`, we make no guarantees on duckdb-httpfs's `main` (and at this moment is also slightly outdated)

carlopi:
And yes, the patches adding Initialize() were the path forward.

carlopi:
Thanks!

PR #807, opened 2025-11-13
Upgrade observefs 0.4.1

dentiny:
This PR upgrades observefs to v0.4.1, the main change is to upgrade duckdb and other related dependencies to compatible version.


carlopi:
Thanks!

PR #812, opened 2025-11-14
Update dns extension to DuckDB v1.4.2

tobilg:


onnimonni:
Hey @tobilg! Thanks for keeping this alive.

I have a question for you. When maintaining this do you always update the version in the MainDistributionPipeline.yml?

I added the versions similiarly as you did into few extensions I created lately:

https://github.com/tobilg/duckdb-dns/blob/main/.github/workflows/MainDistributionPipeline.yml#L20-L21

I'm wondering if it's easier to maintain things when just directly having the main branch there like in the `shellfs`:

https://github.com/Query-farm/shellfs/blob/main/.github/workflows/MainDistributionPipeline.yml#L16-L24

carlopi:
Thanks!

PR #814, opened 2025-11-15
Update otlp community extension to v0.2.0

smithclay:
- Bumps release to `v0.2.0` otlp extension which is built against duckdb v1.4.2
- Minor metadata description updates to clarify purpose and how it integrates with the OpenTelemetry Collector


carlopi:
Thanks!

PR #816, opened 2025-11-16
[WIP] Make observefs windows-compatible

dentiny:


dentiny:
I use a bunch of C++17+ features, which are not available in MSVC (g++/clang usually issues warnings but not compilation failure) :(

PR #818, opened 2025-11-17
Webdavfs: Update to 1.0.1

onnimonni:
I found that if one uses this extension and `sshfs` simultaneously then this extension tries to anyway take over the requests to the Hetzner Storagebox even if the resource prefix is 'ssh://'. This should fix it.


carlopi:
Thanks!

PR #819, opened 2025-11-17
Bump erpl_web to DuckDB v1.4.2

sipemu:
Updated the repository reference to the latest commit.


sipemu:
make test fails as the HTTP request is unreliable. The tests now also accept 502 responses.

carlopi:
Thanks!

PR #820, opened 2025-11-17
Sshfs: Update to 1.0.1

onnimonni:
This fixes major bugs on how libssh2 was used which prevented reading large amount of data from the files.
 
 Now it seems to work for all cases I was able to figure out.


carlopi:
Thanks!

PR #825, opened 2025-11-19
Duckherder upgrade v0.0.3

dentiny:
This PR upgrades duckherder to v0.0.3, which includes a few features (though still experiment, but more play-able):
- Add metrics for distribution mode, query execution time, task/worker number, etc
- Add functions for users to register worker nodes and driver node, which basically allow them to implement them to their own driver/worker, as long as they speak duckherder dialect (i.e., grpc stubs and arrow flight).

Also add Peter into `observefs` extension maintainer.

### DIFF
--- a/extensions/duckherder/description.yml
+++ b/extensions/duckherder/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: dentiny/duckdb-distributed-execution
-  ref: ccb24330caffab48b5c4da62004da9f0bfabd03f
+  ref: 4a03f33228eac4748172bea5d6834b6839e431ce
 
 docs:
   hello_world: |


dentiny:
> Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: pyyaml in /usr/lib/python3/dist-packages (6.0.1)
Files changed: ['extensions/duckherder/description.yml', 'extensions/observefs/description.yml']
Traceback (most recent call last):
  File "/home/runner/work/community-extensions/community-extensions/scripts/build.py", line 14, in <module>
    raise ValueError('cannot have multiple descriptors changed or packages with spaces in their names')
ValueError: cannot have multiple descriptors changed or packages with spaces in their names

not sure what's going on, but maybe it could be related to I'm changing two extensions in one PR? I could try split them though

carlopi:
Thanks!

PR #829, opened 2025-11-19
Update `lsh` extension to `v0.2.0`

yoonspark:
Changes include:

* Relicensed under MIT to facilitate broader reuse
* Added Jaccard similarity function with an `ngram_width` argument
* Overloaded MinHash functions to enable users to hash their own shingles pre-generated based on their unique needs (e.g., word bigrams with particular stemming)

Thanks for your help!


carlopi:
Thanks!

PR #838, opened 2025-11-25
[PDAL] New extension that adds support for manipulating point cloud data using SQL.

ahuarte47:
This is an extension for DuckDB for manipulating point cloud data using SQL.

You can use the extension to read point cloud data from various formats (e.g., LAS, LAZ) and perform transformations on them.

The extension is built on top of [PDAL (Point Data Abstraction Library)](https://pdal.io/), a C++ library that enables users to read, write, and process point cloud data and, with this extension, load the data directly into DuckDB using SQL queries.


carlopi:
Thanks!!

ahuarte47:
Thanks to you @carlopi !!

PR #843, opened 2025-11-26
Upgrade cache httpfs v0.9.4

dentiny:
This PR upgrades cache httpfs to v0.9.4, two main changes:
- Users reports empty local cache files, so I added debug logging message for local file deletion and creation
- Add metrics for bytes to read and bytes to cache, which could be used to decide cache block size


carlopi:
Thanks!

PR #849, opened 2025-11-29
Add macos support for duckherder

dentiny:
This PR adds macos support for duckherder extension, mainly two changes:
- I addressed some flaky local server startup issues, and checked it works on my macos development environment.
- Downgrade the C++ compilation version from 20 to 17, the issue lies in (1) arrow uses forward declared type in vector; but (2) libc++ (clang's default standard library on macos) decorates constexpr, which requires full definition.

### DIFF
--- a/extensions/duckherder/description.yml
+++ b/extensions/duckherder/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: dentiny/duckdb-distributed-execution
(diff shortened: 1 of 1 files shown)


dentiny:
Should be good to review now, the issue lies in C++20/arrow-cpp, I downgrade the compiler version to 17.

carlopi:
Thanks!!

PR #850, opened 2025-11-30
Add http_request extension

onnimonni:
## Summary

This is very similiar compared to the existing https://duckdb.org/community_extensions/extensions/http_client but this also:

- supports HEAD/GET/POST/PUT/PATCH/DELETE methods
- Can do byte-range requests with `byte_range()` helper
- Auto-decompresses gzip/zstd responses
- Supports the http proxy configs with the same proxy flags and proxy secrets as httpfs and httputil.


carlopi:
Thanks!

PR #852, opened 2025-11-30
Add warc extension

onnimonni:
## Summary

Add the `warc` extension for parsing WARC (Web ARChive) records, the standard format used by Common Crawl and web archiving tools.

**Features:**
- `parse_warc(BLOB|VARCHAR)` - Parse WARC records and return structured data
- Auto-detects gzip compression
- Returns struct with: warc_version, warc_headers, http_version, http_status, http_headers, http_body
- HTTP body returned as BLOB for binary safety
- HTTP header keys lowercased for consistent JSON access

**Use case:**
Process Common Crawl data efficiently by fetching only specific byte ranges and parsing them directly in DuckDB.

**Example:**
```sql
-- Parse a WARC record
SELECT
    (parse_warc(content)).http_status,
    decode((parse_warc(content)).http_body) as html
FROM read_blob('record.warc.gz');
```

Repository: https://github.com/midwork-finds-jobs/duckdb_warc

### DIFF
--- a/extensions/warc/description.yml
+++ b/extensions/warc/description.yml
@@ -1,0 +1,81 @@
+extension:
+  name: warc
+  description: Parse WARC (Web ARChive) records for Common Crawl data processing
+  version: 0.1.0
+  language: Rust
+  build: cargo
+  license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #857, opened 2025-12-02
Add yardstick extension

nicosuave:
Measure-aware SQL implementing Julian Hyde's "Measures in SQL" paper.

### DIFF
--- a/extensions/yardstick/description.yml
+++ b/extensions/yardstick/description.yml
@@ -6,7 +6,7 @@ extension:
   build: cmake
   license: MIT
   requires_toolchains: "rust"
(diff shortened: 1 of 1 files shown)


carlopi:
There is something very weird going on on the MacOS amd build, might very well be on our side / the infrastructure, or maybe some flag is missing while passing to rust.

Could you take a look? Thanks!

nicosuave:
> There is something very weird going on on the MacOS amd build, might very well be on our side / the infrastructure, or maybe some flag is missing while passing to rust.
> 
> Could you take a look? Thanks!

Yeah I reworked it a bit & I think it should pass now 🤞 thanks!

carlopi:
Thanks!

PR #859, opened 2025-12-02
Upgrade sysem stats extension to v0.2.0

dentiny:
This PR upgrades system_stats extension to v0.2.0, which accomplishes all stats I want to expose at the moment.


carlopi:
Thanks!

PR #863, opened 2025-12-03
Add extension workflow split script

samansmink:
This script helps you build all extensions without giving GH actions a heart attack @taniabogatsch


taniabogatsch:
Would I run that script locally or in CI? Since it looks like this PR does not add it to a CI run?

samansmink:
No you run it locally, it just produces a bunch of gh cli commands to run

carlopi:
Thanks!

PR #900, opened 2025-12-06
Netquack 1.7.0

hatamiarash7:
## What's Changed

- Thread safety logger
- Improve IP utils
  - RFC 3021 /31 support
  - Replaced regex with manual character parsing (faster)
  - Prevent overflow
- SQL injection vulnerability: Changed raw string concatenation to parameterized queries using `con.Prepare()`
- Renamed `file` to `download_file` to avoid shadowing
- Returns NULL instead of an empty string when the domain is not found
- Fixed `extract_host` to properly handle `user:pass@host` URLs by looking ahead for `@` when encountering `:`
- Added `constexpr`/`noexcept`
- Improve state handling
- Handle NULL inputs everywhere
- Fixed to return empty when `#` comes before `?` (RFC-compliant)
- Remove unused includes and variables
- Write new tests
- Bump DuckDB version to v1.4.2
- Bump Extension CI Tools to v1.4.3

---

**Full Changelog**: https://github.com/hatamiarash7/duckdb-netquack/compare/1.6.0...1.7.0


carlopi:
Thanks!

PR #904, opened 2025-12-08
Update cache httpfs to v0.10.0

dentiny:
Two main changes:
- Fix a chunk calculation bug, which would lead to an extra unnecessary empty IO request
- Add options to validate in-memory or on-disk cache entries with version tag, by default off


carlopi:
Thanks!

PR #913, opened 2025-12-09
Update duckpgq to v1.4.3

Dtenwolde:
Think this should be good to go, just need to verify that this completes: https://github.com/cwida/duckpgq-extension/actions/runs/20068182076


carlopi:
Thanks! There seems to be a problem uncovered by one of the tests

Dtenwolde:
Interesting! Will investigate :)

carlopi:
Thanks!

Issue #935, opened 2025-12-09
Windows Arm Extensions


rustyconover:
Please build community extensions for the Windows ARM platform.

Thanks,

Rusty

PR #936, opened 2025-12-09
Support opt-in platforms

staticlibs:
This PR adds support for `opt_in_platforms` attribute to the extension's `description.yml` file. This attribute is optional, it can contain the semicolon-separated list of platforms that have `"opt_in": true` set in the `distribution_matrix.json`, for example: 'windows_arm64;'. It is handled the same way as existing `excluded_platforms` attribute.

Additionally `more_opt_in` input parameter is added to `build.yml` workflow, it is handled the same way as the existing `more_excluded` parameter and is intended to be used to test new opt-in architectures.

Fixes: #935

### DIFF
--- a/extensions/quack/description.yml
+++ b/extensions/quack/description.yml
@@ -14,6 +14,8 @@ extension:
   vcpkg_commit: "ce613c41372b23b1f51333815feb3edd87ef8a8b"
   # (Optional) this extension requires additional custom toolchain setup
   custom_toolchain_script: true
+  # (Optional) ';' separated list of additional platforms
+  opt_in_platforms: "windows_arm64;"
 
 repo:
   github: duckdb/extension-template


rustyconover:
Hi @staticlibs,

I’d really prefer if we didn’t need to manually opt in and out. Having to opt in all 29 extensions would end up costing a lot of CI time.

Would it be possible to have all platforms opted in for all extensions by default, and then 
selectively opt out as needed instead? That would simplify things a lot.

Thanks,
Rusty

rustyconover:
I'm just trying from a perspective of trying to do the least work possible for new platforms when code comes out.

I just want you guys to try to build everything, and if its breaks let me know and I'll decide if I want to opt-out or fix it.

Rusty

rustyconover:
I am pro new platforms.

staticlibs:
@rustyconover 

Hi! I think your request is perfectly valid, just not sure what approach is better here, perhaps the added `more_opt_in` workflow parameter is enough for that: CC @carlopi , @samansmink .

staticlibs:
@carlopi 

Added `windows_arm64` opt-in to quack extension.

carlopi:
Thanks!

PR #944, opened 2025-12-10
Update dns and fakeit extensions to the newest DuckDB v1.4.3

tobilg:


carlopi:
Hey, we currently we are limiting a single extension per PR, could you file separate ones? thanks

PR #959, opened 2025-12-11
Make the build workflow work with DuckDB `v1.4.3`

habedi:


habedi:
The build passed yesterday, but it failed today. Is this a transient error?

carlopi:
Thanks!

PR #966, opened 2025-12-12
Update html_query to v0.1.1

onnimonni:
## Summary
- Add BLOB input support for all functions
- New extract parameter API (replaces text_only boolean)
- html_query_all now returns VARCHAR[] instead of JSON
- Multi-attribute extraction support

## Changes
- version: 0.1.0 → 0.1.1
- ref: updated to latest commit


carlopi:
Thanks!

PR #969, opened 2025-12-12
Add: duck_hunt

teaguesterling:
Adding duck_hunt, a general purpose log and result parser for lots of different types of linters, build systems, CI workflows, etc.


carlopi:
Thanks!

PR #972, opened 2025-12-14
add: sitting_duck

teaguesterling:
Sitting duck allows duckdb to wrap tree-sitter AST parsing framework to query syntax trees (and add normalization) for 20+ languages including a special handler for DuckDB's internal parser.


carlopi:
Thansk!

PR #976, opened 2025-12-14
Bump up cache httpfs to v0.11.0

dentiny:
Two main changes:
- [fix] Allow one single process to run multiple duckdb instances with cache httpfs extension
- [improvement] Leverage duckdb setting callback to update instance state config active, instead of check all configs passively at every IO operation


carlopi:
Thanks!

PR #979, opened 2025-12-14
sitting_duck: bump to v1.0.1 - fix segfault

teaguesterling:
Fix segfault at extension load from lingering old extension API


teaguesterling:
Sorry again for the spam @carlopi - I had left a remnant of the old extension API in place that caused a segfault when actually loading outside of the CI chain.

carlopi:
Thanks!

PR #991, opened 2025-12-19
Updates for the rusty_sheet extension v0.4.2

redraiment:
* Build plugin using the STABLE DuckDB C API, ensuring forward compatibility without repackaging on future DuckDB upgrades
* Update DuckDB support to version 1.4.3
* Fix incorrect field type detection when processing nulls in non‑string columns


carlopi:
Thanks!

PR #995, opened 2025-12-19
Update http_request to v0.2.3

onnimonni:
## Summary
- Update http_request extension to v0.2.1
- Adds User-Agent header override support
- Remove unused curl dependency
- Exclude MinGW builds (OpenSSL vcpkg race condition)


carlopi:
Thanks!

PR #996, opened 2025-12-21
Update cache httpfs to v0.11.1

dentiny:
Two main changes:
- Delete cache entries on write
- Implement extended APIs for duckdb filesystem interfaces (open and list)


carlopi:
Thanks!

PR #999, opened 2025-12-21
Update http_request to v0.2.4

onnimonni:
## Summary
- Fix duplicate Set-Cookie headers causing "map keys must be unique" error
- Parse Set-Cookie headers into structured `cookies[]` array
- Add convenience fields: `content_type`, `content_length`

## Changes
The response structure is now:
```sql
STRUCT{
  status: INTEGER,
  content_type: VARCHAR,
  content_length: BIGINT,
  headers: MAP(VARCHAR, VARCHAR),
  cookies: STRUCT{name, value, expires, max_age, path, domain, secure, httponly, samesite}[],
  body: BLOB
}
```

## Test plan
- [x] Local build passes
- [x] Tests pass

### DIFF
--- a/extensions/http_request/description.yml
+++ b/extensions/http_request/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: http_request
   description: HTTP client extension for DuckDB with GET/POST/PUT/PATCH/DELETE and byte-range requests
-  version: 0.2.3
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1000, opened 2025-12-22
Update http_request to v0.3.0

onnimonni:
- Add parallel HTTP execution for scalar functions
- Requests within a chunk now run concurrently (up to 32 by default)
- New http_max_concurrency setting to control parallelism


carlopi:
Thanks!

PR #1003, opened 2025-12-22
Bump `minijinja` to latest commit

rustyconover:


carlopi:
@rustyconover: the diff is empty, so this has no effect

rustyconover:
I think I have an automation problem on my end, will resolve.

PR #1023, opened 2025-12-29
Update cache httpfs to v0.11.2

dentiny:
bump up version, which contains a hot patch to disable cached compressed read, need long-term solution.


carlopi:
Thanks!

PR #1031, opened 2025-12-31
Upgrade anndata extension to v0.11.5

honicky:
## Summary

Upgrade the AnnData extension to v0.11.5.

### Bug fixes in v0.11.5
- Fixed reading obs/var counts from X matrix shape attribute when _index datasets don't exist
- Fixed bug where shape attribute was incorrectly read from file root instead of /X group for sparse matrices

### Bug fixes in v0.11.4
- Fixed categorical columns with int16/int32 codes (>127 categories)
- Fixed integer and float categorical columns in obs/var tables

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,12 +93,12 @@ jobs:
 
   build:
     needs: prepare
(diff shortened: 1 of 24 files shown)


carlopi:
Thanks!

PR #1032, opened 2026-01-01
scalarfs: new extension

teaguesterling:
Adding an extension that adds a few utility filesystems like `variable:` and `data:`.


carlopi:
Amazing, thanks!

PR #1057, opened 2026-01-06
duck_hunt: bump to v1.3.1

teaguesterling:
- **Transparent compression support** - Automatically decompress `.gz` and `.zst` files based on file extension
  - GZIP (`.gz`) is built-in and always available
  - ZSTD (`.zst`) requires `LOAD parquet` first


carlopi:
Thanks!

PR #1059, opened 2026-01-06
duck_block_utils: bump to v1.2.0

teaguesterling:
Bump duck_block_utils from v1.0.0 to v1.2.0.

Changes since v1.0.0:
- `pandoc_ast()` table function with meta/api_version params
- Nested list, div, blockquote, table, image Pandoc conversions
- Skip metadata blocks in Pandoc output
- 474 test assertions


carlopi:
Thanks!

PR #1068, opened 2026-01-08
http_request: v0.5.0 - add http_user_agent setting

onnimonni:
## Summary
- Add `http_user_agent` setting for global User-Agent override

## Usage
```sql
SET http_user_agent = 'MyApp/1.0';
SELECT http_get('https://api.example.com/');
```

Per-request headers still override the global setting.


carlopi:
Thanks!

PR #1076, opened 2026-01-11
Update http_request to v0.6.0

onnimonni:
## Changes

- Add request caching to prevent duplicate HTTP requests within same query
- Fix crash when using MAP syntax for headers (`headers := MAP {'key': 'value'}`)
- Headers now support both STRUCT and MAP syntax

## New settings

- `http_request_cache` (default: true) - Cache HTTP responses within query execution to prevent duplicate requests

## Bug fixes

- Fixes crash when using `headers := MAP {...}` syntax (was only supporting STRUCT)


carlopi:
Thanks!

PR #1078, opened 2026-01-11
markdown: bump to v1.3.9

teaguesterling:
Fix WASM compatibility issue (teaguesterling/duckdb_markdown#13) and Windows frontmatter regex matching bug.

**v1.3.8**: WASM fix - cmark-gfm function pointer casts caused indirect call signature mismatch errors in WebAssembly.

**v1.3.9**: Windows fix - MSVC regex `^` anchor wasn't correctly matching only at the start of the string, causing false positive frontmatter detection.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/extensions/markdown/description.yml
+++ b/extensions/markdown/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: markdown
   description: Read, analyze, and write Markdown files with block-level document representation and inline element support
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1083, opened 2026-01-12
Upgrade observefs to v0.4.5

dentiny:
Upgrade observefs to v0.4.5, which records most of the IO operations latency.


carlopi:
Thanks!

PR #1087, opened 2026-01-13
Fix pre-release deploy

samansmink:
This should resolve the issue where community extensions builds targeting a non-existent release (as is done in duckdb's release process) are uploaded to the versioned path instead of the tag-based path


carlopi:
Is any change needed in how to community-extensions builds are triggered?

samansmink:
@carlopi I think not, we were already setting the tag. I will double check this by test-running it on `quack` after this is merged

PR #1088, opened 2026-01-13
Bump boilstream to v0.5.0

dforsber:
- Matches with `boilstream-0.8.0`
- New table function `boilstream_buckets()` that lists available S3 Buckets to create DuckLakes with


carlopi:
Thanks!

PR #1090, opened 2026-01-14
Update otlp extension to v0.3.0

smithclay:
- Bumps to OTLP extension v0.3.0
- Builds against duckdb v1.4.3
- Adds support for reading protobuf files in wasm
- Uses new standardized schema used in other OTLP projects
- Fairly major refactor: core extension code now uses https://github.com/smithclay/otlp2records **
- Drops support for some wasm platforms and linux musl (not supported by our new shared rust crate)

** This isn't a pure rust extension (yet) to preserve support for wasm_eh builds


carlopi:
Thanks!

PR #1093, opened 2026-01-14
Update http_request to v0.8.0

onnimonni:
## Changes
- Add `http_follow_redirects` setting to disable auto-follow redirects (default: true)
- Add `HTTP_HEADERS` type with case-insensitive `[]` access

### Redirect Control
```sql
SET http_follow_redirects = false;
SELECT status, headers['Location'] FROM http_get('https://example.com/redirect');
-- Returns 301/302 with Location header instead of following
```

### Case-Insensitive Header Access
HTTP headers are case-insensitive per RFC 7230. The new `HTTP_HEADERS` type provides native case-insensitive `[]` operator:

```sql
-- All return the same value
SELECT headers['Content-Type'] FROM http_get('https://example.com/');
SELECT headers['content-type'] FROM http_get('https://example.com/');
SELECT headers['CONTENT-TYPE'] FROM http_get('https://example.com/');
```

### DIFF
--- a/extensions/http_request/description.yml
+++ b/extensions/http_request/description.yml
@@ -11,7 +11,7 @@ extension:
 
 repo:
   github: midwork-finds-jobs/duckdb_http_request
-  ref: 069e6ff997673daeda30fd6180887c459fcdfc30
+  ref: e55035fe799fc7173fa8b3706e1d88c866a048d3
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


onnimonni:
The build failures are in httpfs extension, not http_request:

```
/duckdb_build_dir/build/release/_deps/httpfs_extension_fc-src/src/httpfs.cpp:611:17: error: 'DUCKDB_LOG_WARNING' was not declared in this scope; did you mean 'DUCKDB_LOG_WARN'?
```

This appears to be a community-extensions infrastructure issue - httpfs using outdated DuckDB API. http_request builds and tests pass locally.

carlopi:
Thanks!

onnimonni:
@carlopi sorry to bother you but any chance to get opinion on this: https://github.com/duckdb/duckdb/pull/20460

This extension is currently using the duckdb internal cpp-httplib. I can replace it with curl but then I would need to integrate all http and proxy settings separately because now I can just rely on the existing settings from: https://duckdb.org/docs/stable/core_extensions/httpfs/https.

This makes it much cleaner when user adds authentication headers or proxy settings or custom certificate files and they will work both on the default duckdb `FROM 'https://example.com/internal-api.json'` and with the http_X methods from this extensions.

PR #1100, opened 2026-01-18
markdown: bump to v1.3.10

teaguesterling:
Another attempt at fixing the function pointer casts in cmark-gfm for WASM compatibility (ref: https://github.com/teaguesterling/duckdb_markdown/issues/13)


carlopi:
Thanks!

PR #1109, opened 2026-01-20
Update crawler extension repo name

onnimonni:
Update repository name from `duckdb-crawl` to `duckdb-crawler` after rename.


carlopi:
Thanks!

PR #1165, opened 2026-01-30
Add Elasticsearch extension

tlinhart:


carlopi:
Thanks! And welcome!

This will be built and deployed, to be available like `INSTALL elasticsearch FROM community;` then later today the docs will also be up.

PR #1188, opened 2026-02-03
Remove encoding extension because it's not maintained anymore

onnimonni:


taniabogatsch:
Hi @onnimonni - thanks for the PR. Let me just discuss how we usually go about community extension deprecation and then get back to you. 👍

onnimonni:
Hey!

This was my test of trying to learn how to do these extensions in the first place. I don't think it's that useful for anyone in reality and thus I wanted to simplify the community extensions a bit.

carlopi:
I think it's correct that `encodings` has been moved to core repository, but I think we would need a 'deprecated' path or so, allowing to encode the fact that a given extension was at some point existing.

I would need to check how other package manager do this, do you have any idea?

carlopi:
Thanks!

PR #1197, opened 2026-02-04
submit v0.1.1

vhe74:
Dqtest is a framework to allow data-testing directly inside DuckDB


carlopi:
Thanks! And welcome!

This should build / be available in 1 hour or so, docs later today.

vhe74:
Thank you Carlo !

PR #1244, opened 2026-02-11
read_lines: new extension

teaguesterling:
This is a very simple extension that allows for selective extraction of lines by line number from files. Could easily be done in SQL bu this prevents materializing large amounts of unused data and allows providing explicit line numbers.

### DIFF
--- a/extensions/read_lines/description.yml
+++ b/extensions/read_lines/description.yml
@@ -10,7 +10,7 @@ extension:
 
 repo:
   github: teaguesterling/duckdb_read_lines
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

teaguesterling:
@carlopi sorry for the noise: but is there a reason this isn't available yet? The other update from the same time did deploy, but I can't install this (new) one.

carlopi:
Thanks for the ping, there was some GH hiccup with networking, and somehow it stalled. Restarted at https://github.com/duckdb/community-extensions/actions/runs/21937884130

PR #1249, opened 2026-02-12
Bump chaos for v1.5.0

taniabogatsch:
Add `next_ref`, which now points to a version that builds for `v1.5-variegata`.
Also bump the current `ref`, since that is now on a separate branch containing some pushes.


carlopi:
Thanks!

PR #1251, opened 2026-02-12
Add level_pivot extension

halgari:
Adds the `level_pivot` extension, that adds leveldb interop to DuckDB along with high performance key/value to table pivot functionality.


carlopi:
Thanks!

Note that testing for `v1.5-variegata` has been enabled, so the failing tests are due to that, but not a blocker.
To provide a commit compatible with upcoming v1.5.0, please add a `ref_next` pointing to the appropriate `sha` in your extension repo

PR #1252, opened 2026-02-13
Bump `hashfuncs` to latest commit

rustyconover:


taniabogatsch:
Thanks!

Note that testing for `v1.5-variegata` has been enabled, so the failing tests are due to that, but not a blocker.
To provide a commit compatible with the upcoming `v1.5.0`, please add a `ref_next` pointing to the appropriate `sha` in your extension repo.

PR #1254, opened 2026-02-13
Bump anofox-tabular to duckdb v1.4.4

sipemu:
Updated the GitHub reference for the repository.

### DIFF
--- a/extensions/anofox_tabular/description.yml
+++ b/extensions/anofox_tabular/description.yml
@@ -3,7 +3,7 @@ extension:
   description: A duckdb extension which combines data quality and data preparation tools for tabular data.
   language: C++
   build: cmake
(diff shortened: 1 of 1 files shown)


sipemu:
The Linux ARM64 build failed because vcpkg couldn't download a dependency of OpenVINO. Maybe a re-run could help.

taniabogatsch:
Yea, we've been running into this a few times now. [I've added a comment here](https://github.com/microsoft/vcpkg/issues/49963) but will also re-run the jobs (again) now.

taniabogatsch:
@sipemu - could you please merge with `main` here? I think that should fix the CI failures.

taniabogatsch:
Restarted one more time now, let's hope...

sipemu:
@taniabogatsch I excluded linux_arm64. It should now work. My fault, I missed that we do not have linux_arm64 in our build pipeline.

carlopi:
Thanks! Feel free to add a v1.5-compatible ref, but this is good to go.

PR #1287, opened 2026-02-13
Enable 1.5 builds for `tsid`

rustyconover:


taniabogatsch:
Re-running the failed jobs because IIRC we've seen that (likely spurious) VCPKG ninja failure in other CI runs before.

taniabogatsch:
Thanks!

PR #1291, opened 2026-02-14
Add extension: hedged request fs

dentiny:
This PR adds a new community extension `hedged_request_fs`, which is used to cut down tail latency for IO operations.


carlopi:
Thanks!

Issue #1299, opened 2026-02-14
build_next.yml: CRT mismatch when testing extensions against DuckDB v1.5


VGSML:
## Problem

`build_next.yml` tests extensions against DuckDB v1.5 (`duckdb_version: v1.5-variegata`) but uses `ci_tools_version: v1.4-andium` (inherited from `build.yml`). This creates a CRT (C Runtime Library) linkage mismatch on Windows MSVC builds.

**DuckDB v1.5** sets `CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded..."` (commit e9f5d4a518, Dec 2025) which means `/MT` — static CRT (`libucrt.lib`).

**extension-ci-tools v1.4-andium** uses triplet `x64-windows-static-md-release-vs2019comp` which sets `VCPKG_CRT_LINKAGE dynamic` — meaning `/MD` (dynamic CRT, `ucrt.lib`).

When vcpkg packages (like OpenSSL) are compiled with `/MD` but linked against DuckDB compiled with `/MT`, the linker produces unresolved `__imp_` symbol errors for CRT functions.

## Failing CI

https://github.com/duckdb/community-extensions/actions/runs/21999261929/job/63591936205

Example linker errors:
```
openssl.lib(libcrypto-lib-o_time.obj) : error LNK2019: unresolved external symbol __imp__time64
openssl.lib(libcrypto-lib-ctype.obj) : error LNK2019: unresolved external symbol __imp_strspn
openssl.lib(libcrypto-lib-ctype.obj) : error LNK2019: unresolved external symbol __imp_strncpy
```

The `__imp_` prefix means "import from DLL" — these symbols exist in `ucrt.lib` (dynamic import lib for `/MD`) but not in `libucrt.lib` (static CRT for `/MT`).

## Root Cause

extension-ci-tools PR #276 (commit cd7a046) already fixed this on `main` by switching the Windows triplet from `x64-windows-static-md-release-vs2019comp` to `x64-windows-static-release` (static CRT matching DuckDB v1.5's `/MT`). However, this fix was **not backported** to the `v1.4-andium` branch.

Since `build_next.yml` inherits `ci_tools_version` from `build.yml` (which correctly uses `v1.4-andium` for DuckDB v1.4.4), the v1.5 test job gets the wrong CI tools version.

## Suggested Fix

Option A: Override `ci_tools_version` in `build_next.yml` to use `main` (or a v1.5-compatible tag) of extension-ci-tools.

[...]

PR #1302, opened 2026-02-14
Allow to provide ci_tools_version as input, and use that for build_next

carlopi:
Not yet tested, but should on paper be simple.

Thanks @VGSML & @rustyconover to report the issue. Should fix https://github.com/duckdb/community-extensions/issues/1299.

### DIFF
--- a/extensions/quack/description.yml
+++ b/extensions/quack/description.yml
@@ -19,7 +19,7 @@ extension:
 
 repo:
   github: duckdb/extension-template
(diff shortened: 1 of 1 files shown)


carlopi:
Note: in the current status there is no testing, since it require the build.yml workflow present in `main` to be aware of a non existent parameter.
To clean that up / allow testing, I'll first land https://github.com/duckdb/community-extensions/pull/1303

PR #1549, opened 2026-03-18
Update Duck Lineage extension to v0.2.0

thijs-s:
Duck Lineage 0.2.0 brought initial support for column lineage gathering, so we decided to make a new release.


carlopi:
Thanks!

PR #1616, opened 2026-03-27
Add HTTP stats extension

tlinhart:
Add new HTTP stats extension which provides better HTTP traffic stats in EXPLAIN ANALYZE output than standard HTTPFS extension.

Extension repo: https://github.com/tlinhart/duckdb-http-stats


carlopi:
Thanks!

PR #1680, opened 2026-04-02
Update sitting_duck to v1.6.0

teaguesterling:
## Summary
- Update sitting_duck extension from v1.5.0 to v1.6.0
- New: `ast_select()` — CSS selector query language bootstrapped with tree-sitter-css
- New: `ast_type_map()` — node type discovery across 27 languages
- New: ~80 semantic type aliases (.func, .if, .except, .import, etc.)
- New: NAME_ROLE flags (is_name_definition, is_name_reference, is_scope)
- New: True parallel file parsing via DuckDB init_local + MaxThreads API
- New: Bare type prefix matching (`if` matches `if` + `if_statement` + `if_clause`)
- Fix: Relaxed wildcard child matching (async/modifier tolerant)
- Fix: sibling_index UINT32 → INT32 (no more unsigned overflow)

## Release
https://github.com/teaguesterling/sitting_duck/releases/tag/v1.6.0

### DIFF
--- a/extensions/sitting_duck/description.yml
+++ b/extensions/sitting_duck/description.yml
@@ -9,7 +9,7 @@ extension:
     - teaguesterling
 repo:
   github: teaguesterling/sitting_duck
-  ref: fa5454689bdd869d997a54bbfa28e8fde166b3a8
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1712, opened 2026-04-09
Bump duckpgq to fix build errors

Dtenwolde:
Attempt to fix: 
```sql 
memory D force install duckpgq from community; 
memory D load duckpgq; 
Invalid Input Error:
Initialization function "duckpgq_duckdb_cpp_init" from file "/Users/danieltenwolde/.duckdb/extensions/v1.5.1/osx_arm64/duckpgq.duckdb_extension" threw an exception: "vector"
```


carlopi:
Thanks!

nicolas-geysse:
Great , thanks ! It seems the binary is not rebuild, yet, on the CDN (not complaining, just trying to understand here)

PR #1715, opened 2026-04-10
New decimal arithmetic extension

Dtenwolde:
This extension adds a scalar function for decimal division. Rather than returning a double like the normal division, it return a decimal with a certain precision and scale. It also adds a decimal average function and a couple of rounding functions all returning decimals.


carlopi:
I think there might be an issue with naming?

`decimal-arithmetic` (extension folder) vs `decimal_arithmetic` (source code)

dentiny:
image
Hmmm I thought the inconsistency should be detected at prepare stage after https://github.com/duckdb/community-extensions/pull/1531, will check.

----

Update:
> ValueError: Directory name 'decimal_arithmetic' does not match extension name 'decimal-arithmetic' in extensions/decimal_arithmetic/description.yml. Please rename the directory to 'decimal-arithmetic'.

Seems to work fine haha

Dtenwolde:
> Update:

>    ValueError: Directory name 'decimal_arithmetic' does not match extension name 'decimal-arithmetic' in extensions/decimal_arithmetic/description.yml. Please rename the directory to 'decimal-arithmetic'.

> Seems to work fine haha

let me try!

carlopi:
Thanks!

PR #1719, opened 2026-04-11
Update sitting_duck to v1.7.0

teaguesterling:
Major release bringing three new user-visible features, one format change, and a large round of internal ast_select refactoring.

Release notes: https://github.com/teaguesterling/sitting_duck/blob/v1.7.0/docs/releases/v1.7.0.md

## Highlights

**`parse_ast_list()` scalar** ([#62](https://github.com/teaguesterling/sitting_duck/issues/62)) — `LIST<STRUCT>` variant of `parse_ast()` usable inside CTEs, scalar expressions, and lateral joins where a table function isn't allowed. Internally used by `ast_select`'s pattern-matching pseudo-classes.

**`+schema` extraction suffix** ([#61](https://github.com/teaguesterling/sitting_duck/issues/61)) — any of `context`/`source`/`structure`/`peek` can take a `+schema` suffix that keeps the full column set in the output schema as NULLs without computing them. Enables SQL macros to declare stable schemas independent of which columns they actually populate. `ast_select` uses `peek := 'none+schema'` internally.

**`qualified_name` format overhaul** — changed from `L/NAME` to `L[NAME]` with self-delimiting brackets, plus a collision-disambiguating `[N]` suffix for same-scope duplicates:

```
counter = 0   -> V[counter]
counter = 1   -> V[counter][2]
counter = 2   -> V[counter][3]
```

`qualified_name` is now **unique within a file** for all named definition nodes. The bracket format enables clean LIKE matching (`LIKE '%F[%'` for all functions, `LIKE '%[foo]%'` for name "foo") without false positives.

**`:match` / `:contains` pseudo-classes** (replacing `:matches`) — CSS's `:matches`/`:is` historically operates on the current element, but sitting duck's `:matches` was doing a subtree scan. Now fixed:

```sql
-- Strict: target must be a call node
SELECT * FROM ast_select('src/*.py', 'call:match("db.execute()")');

-- Lenient: function whose subtree contains the pattern (old :matches behavior)
SELECT name FROM ast_select('src/*.py', '.func:contains("db.execute()")');
```

## Breaking changes

- **`qualified_name` format `L/NAME` → `L[NAME]`**. Any `LIKE` patterns or equality comparisons against literal qualified_name strings need updating. See the release notes for a migration table.
- **`:matches()` removed** (not aliased). Replace with `:contains()` for identical semantics.

## Carried-forward v1.6.1 fixes

v1.6.1 was tagged for internal history but not published to community — its fixes ship in v1.7.0:
- `:not(:pseudo-class)` now actually filters (previously silently dropped)
- Unknown pseudo-classes raise clear errors instead of silently matching everything

## Test plan

- [x] All 99 existing tests pass (4786 assertions, up from 93 / 4671 at v1.6.0)
- [x] New `test/sql/schema_suffix.test` for `+schema` behavior
- [x] New `test/sql/parse_ast_list.test` for the scalar function
- [x] New `test/sql/qualified_name.test` sections 16-22 for collision suffix semantics
- [x] Updated `test/sql/ast_select_native.test` for `:match` / `:contains` semantics
- [x] Updated existing `qualified_name.test` assertions for the `L[NAME]` format

### DIFF
--- a/extensions/sitting_duck/description.yml
+++ b/extensions/sitting_duck/description.yml
@@ -9,7 +9,7 @@ extension:
     - teaguesterling
 repo:
   github: teaguesterling/sitting_duck
-  ref: 6d3386f1e9768c8eb5b4abf6752f57a39a8721ef
+  ref: 21e47fe226f368bb57da5e58ae74f53d0ec6ba48
 docs:
   hello_world: |
     -- Parse Python code and find function definitions
--- a/extensions/sitting_duck/description.yml
+++ b/extensions/sitting_duck/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: sitting_duck
   description: Parse and analyze source code ASTs from 27 programming languages with tree-sitter grammars, pattern matching, and structural search
-  version: 1.6.0
+  version: 1.7.0
   language: C++
   build: cmake
   license: MIT
@@ -9,7 +9,7 @@ extension:
     - teaguesterling
 repo:
   github: teaguesterling/sitting_duck
-  ref: 908927e016cbe01f0786582fb00cf8e0de9c6009
+  ref: 6d3386f1e9768c8eb5b4abf6752f57a39a8721ef
 docs:
   hello_world: |
     -- Parse Python code and find function definitions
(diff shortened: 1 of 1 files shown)


teaguesterling:
Moving to draft while we evaluate a potential qualified_name format revision. Don't merge yet — I'll re-ready this once the format question is settled.

teaguesterling:
Ready for review. Revised qualified_name design since the previous draft state: the column is now `LIST<STRUCT<semantic_type SEMANTIC_TYPE, name VARCHAR, index INTEGER>>` (structured scope path) rather than the interim bracket-string form. See the updated release notes at docs/releases/v1.7.0.md for the full story and migration notes. All 98 tests / 4788 assertions pass against ref `21e47fe`.

carlopi:
Thanks!

PR #1723, opened 2026-04-11
maxmind: v0.9.0 Shield/Precision db type support

marselester:
Changes:
- decoding is skipped on projection pushdown
- support of Shield and Precision database type variants
- prefix length when parsing a network is fixed


carlopi:
Thanks!

marselester:
Thank you!

PR #1734, opened 2026-04-14
Add query condition cache extension

dentiny:
This PR introduces query condition cache extension.
CC @Andrewtangtang / @peterxcli


dentiny:
Somehow the CI is not triggered, draft/undraft, open/close PR doesn't work; I'm wondering if team could help unblock? Thanks!

carlopi:
I think it would need to be `description.yml`, and we could use a way to point this to contributors.

dentiny:
> I think it would need to be description.yml, and we could use a way to point this to contributors.

Ahhh sorry my bad

carlopi:
Thanks!

PR #1771, opened 2026-04-15
Bump `decimal_arithmetic`

Dtenwolde:
Changed the behaviour of `decimal_avg` to make the output type be equal to the input type (Used to default to width of 38, and scale based on input)


carlopi:
Thanks!

PR #1776, opened 2026-04-16
MIINT: Additional WASM tweaks, locally tested

wasade:
Extension now verified as loading locally with Node. I hope this unsticks it for real.


carlopi:
Thanks!

PR #1738, opened 2026-04-14
bump zeek extension to v0.4.0

ynadji:
this adds an option `ignore_file_errors` that silently ignores corrupt/malformed files. thanks @keithjjones!

PR #1778, opened 2026-04-16
bump zeek to 0.4.1

ynadji:
on 0.4.0 my extension was pinned to 1.5.1 still. this bumps it to 1.5.2. 

is that why i'm [not seeing the newer version](https://github.com/duckdb/community-extensions/pull/1738#issuecomment-4260337282)? still figuring out the process here.


carlopi:
Thanks!

PR #1800, opened 2026-04-19
Update dplyr community extension descriptor

mrchypark:
## Summary
- update the `dplyr` community extension descriptor to the current `libdplyr` submission snapshot
- mark the extension as a mixed Rust/C++ build with a Rust toolchain requirement
- exclude `windows_amd64_rtools`, which is not part of this extension's supported toolchain path
- point `repo.ref` to the tested `libdplyr` submission commit

## Notes
- `libdplyr` uses a Rust core with a C API bridge and a DuckDB C++ extension adapter.
- The community descriptor is targeted at the current latest stable DuckDB line.
- Repository CI still keeps a separate DuckDB `1.4.0` compatibility lane, but this should be read as a source compatibility check rather than a claim that one distributed unstable/C++ extension binary spans both `1.4.x` and `1.5.x`.
- The extension exposes parser-extension syntax such as `%>%` and embedded pipelines `(| ... |)`, so the `extended_description` was tightened to explain that behavior more directly.

## Upstream source snapshot
- `mrchypark/libdplyr@01961b7ebf2888e0f29e18c0eb7154dde027e422`

### DIFF
--- a/extensions/dplyr/description.yml
+++ b/extensions/dplyr/description.yml
@@ -19,7 +19,7 @@ extension:
 
 repo:
   github: mrchypark/libdplyr
-  ref: 01961b7ebf2888e0f29e18c0eb7154dde027e422
+  ref: v0.3.3
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1822, opened 2026-04-23
Update dash for DuckDB v1.5.2

gropaul:
Hello, this is a both an update with new functionallity and compatibility with v1.5.2


carlopi:
Thanks!

PR #1827, opened 2026-04-25
Update duck_dggs extension description and version; add IGEO7/Z7 operations

am2222:
Revise the duck_dggs extension description and increment the version to 0.1.6. Introduce IGEO7/Z7 bit-level operations for enhanced functionality, allowing direct manipulation of the packed 64-bit Z7 index.


carlopi:
Thanks!

PR #1847, opened 2026-04-30
Add ggsql extension

thomasp85:
The ggsql extension adds support for the ggsql visualization framework (https://ggsql.org). It provides a new reader that interacts directly with the running DuckDB session, and provides a range of different ways to get your plot:

* Visualisations served by an embedded web server and opened in the default browser
* The Vega-Lite spec returned as a 1x1 table
* A self-contained html returned as a 1x1 table


carlopi:
Thanks! And welcome!

PR #1850, opened 2026-05-01
[mongo] Bump to v0.2.1

### LINKED ISSUE #30: [Bug]: High level fields not queryable?
## Description
In the screenshot below, the encircled fields do not appear to be visible or available to duckdb-mongo, resulting in the error seen in the second screenshot.

Image

Image

## Expected Behavior
Fields available in the database, visible and interactable within compass, would, within reason especially for nested data structures, be able to returned via query in duckdb-mongo. Only these two fields highlighted are not returning... possibly due to their position? I don't know, that's why I'm asking for help.

## Actual Behavior
All data fields except the two encircled are queryable and interactable via duckdb-mongo

## Environment
- DuckDB version: 1.5.1
- MongoDB version: unknown
- Operating System: Win 11 Enterprise
- Extension version (if known): 968cb84

stephaniewang526:
- Bump mongo extension from v0.2.0 to v0.2.1
- Uses random `$sample` for schema inference to detect sparse/optional fields (fixes https://github.com/stephaniewang526/duckdb-mongo/issues/30)
- Merges struct fields across LIST(STRUCT) types during type conflict resolution
- Adds compatibility with DuckDB main

### DIFF
--- a/extensions/mongo/description.yml
+++ b/extensions/mongo/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: mongo
   description: Integrates DuckDB with MongoDB, enabling direct SQL queries over MongoDB collections without exporting data or ETL
-  version: 0.2.0
+  version: 0.2.1
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1868, opened 2026-05-02
bump h5db

jokasimr:
* Use pushed down filters to filter globbed files in `h5_tree` and `h5_ls`
* Error messages include filename of file that triggered error


carlopi:
Thanks!

PR #1870, opened 2026-05-03
Updates Ducktinycc to version  0.2.0

sounkou-bioinfo:
This submission updates the ducktinycc community extension descriptor documentation to mention the new scalar UDF `stability` keyword.  This registers the UDF as volatile for side-effectful cases.


carlopi:
Thanks!

PR #1871, opened 2026-05-03
Update GitHub reference in description.yml

nkwork9999:
The previous ref shipped without the bundled model because tiny_model_data.cpp
is .gitignored (131MB, over GitHub's per-file limit). The new commit ships the
22MB .onnx in assets/ and has CMake auto-embed it at configure time, so the
CI build now produces a binary where pic_bundled_info() reports the model and
pic_embed() works zero-setup as advertised in description.yml.


carlopi:
Thanks!

PR #1884, opened 2026-05-06
Add DICOM extension: Import medical imaging data directly into DuckDB

nmontesg:
# Summary

Functionality to import medical imaging data (DICOM, Digital Imaging and Communication in Medicine) directly into DuckDB. It uses the industry standard [DCMTK library](https://dicom.offis.de/dcmtk.php.en) to read DICOM files and convert them into JSON format.

# Quickstart

```sql
INSTALL dicom FROM community;
LOAD dicom;

-- read one file
FROM read_dicom('path/to/dicom_file.dcm');
```

# Functions

`read_dicom(filepath[, load_pixel_data=false])` : Read DICOM files directly into DuckDB, return one row per file with columns `path (VARCHAR)` (path to the file) and `dicom_content (JSON)` (JSON-formatted contents of the DICOM file). Supports glob patterns.

# Platforms

WASM targets excluded (wasm_mvp, wasm_eh, wasm_threads) due to incompatibility with third-party DCMTK

### DIFF
--- a/extensions/dicom/description.yml
+++ b/extensions/dicom/description.yml
@@ -1,0 +1,49 @@
+extension:
+  name: dicom
+  description: |
+    DuckDB integration with medical imaging data (DICOM - Digital Imaging and Communication in
+    Medicine).
+  version: 1.0.0
+  language: C++
+  build: cmake
+  licence: MIT
+  maintainers:
+    - nmontesg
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks! And welcome!

PR #1898, opened 2026-05-10
Add gpudb — GPU-accelerated DuckDB on CUDA + Apple Silicon Metal

singhpratech:
Adds the **gpudb** community extension (now at v0.1.3).

## What it does
gpudb registers `gpu_sum`, `gpu_min`, `gpu_max` aggregate operators that DuckDB transparently dispatches to the GPU. Two backends share one codebase:
- **NVIDIA CUDA** (Linux, sm_70+): on-device aggregation with PCIe-amortized transfer
- **Apple Silicon Metal** (macOS, M1+): unified memory, zero-copy

A hybrid CPU/GPU planner picks the backend per cardinality regime. CPU fallback is automatic when no GPU is present.

## What's new in v0.1.3 (since the original PR)
**Hybrid Metal GROUP BY** that auto-dispatches between two paths per query:
- **slot-lock path** — 32,768 hash partitions × 1024 threadgroup-memory slots; 32-bit CAS protects non-atomic 64-bit sum updates (workaround for Apple Silicon's missing 64-bit `atomic_fetch_add`). Sweet spot: `expected_groups ∈ [1024, 16M]`.
- **radix-opt path** — vectorized 4× `ulong4` loads, in-block 8-bit multi-split scatter via simdgroup prefix-sum. For very low or very high cardinality.

Env override: `GPUDB_METAL_GROUPBY_PATH=slotlock|radix` to force a path.

This flipped the most-attacked workload — TPC-H SF10 GROUP BY `l_orderkey` at 15M unique groups — from CPU 1.78× faster (v0.1.2) to **Metal 1.30× faster**.

## Why it's interesting
- First published SQL execution engine targeting Apple Silicon GPUs. Sirius (CIDR 2026, UW + NVIDIA) and cuDF are CUDA-only.
- Hybrid planner with empirical cardinality thresholds — the open problem from Rosenfeld/Breß CSUR 2022 and Cao SIGMOD 2024.

## Verified numbers — Apple Silicon (M4 Max) vs DuckDB CPU 16-thread

The DuckDB CLI uses 16 threads by default on M4 Max — these comparisons are vs that real default, not single-thread. Each row is pinned to a specific BENCHMARK.md cell; no cross-cell glue.

| Workload | DuckDB CPU mt | Metal v0.1.3 | Speedup |
|---|---:|---:|:---:|
| TPC-H SF10 multi-agg fusion `l_quantity` | 27 ms | **1.06 ms** | **25.5×** |
| TPC-H SF10 multi-agg fusion `l_extendedprice` | 26 ms | **1.18 ms** | **22.0×** |
| TPC-H SF10 multi-agg fusion `l_orderkey` | 11 ms | **1.13 ms** | **9.7×** |
| SF10 SUM `l_quantity` HOT | 5 ms | **1.16 ms** | **4.3×** |
| TPC-H SF10 GROUP BY `l_extendedprice` (1.35M unique) | 113 ms | **28.9 ms** | **3.9×** |
| 500M × 1M GROUP BY synthetic | 820 ms | **242 ms** | **3.4×** |
| 1B × 1M GROUP BY synthetic | 2500 ms | **770 ms** | **3.2×** |
| 1B int64 SUM HOT | 40 ms | **16.2 ms** | **2.6×** |
| TPC-H SF1 GROUP BY `l_orderkey` (1.5M unique) | 8 ms | **5.71 ms** | **1.40×** |
| TPC-H SF10 GROUP BY `l_orderkey` (15M unique) | **56 ms** | **42.95 ms** | **1.30×** |
| TPC-H SF10 GROUP BY `l_quantity` (50 unique) | 26 ms | 372 ms | CPU 14× — structural |

9 wins, 1 honest loss documented. The loss is structural — the CPU's hash table fits in L1 cache for ≤ 1K unique groups; no GPU implementation can beat it, and the auto-dispatcher routes those workloads to the radix path anyway.

## Verified numbers — NVIDIA RTX 4090 (vs single-thread CPU baseline)
- Peak SUM ratio: ~22.8× (size-dependent, resident column)
- GROUP BY 50M rows × 10M unique groups: **21.8×**
- TPC-H SF1 GROUP BY (6M × 1.5M unique): **3.56×** (re-bench post-launch — supersedes the original 4.8× claim)

## Source / details
- Repo: https://github.com/singhpratech/duckdbgpumetaldbram
- Pinned ref: **v0.1.3** = `3ebcd764d28d16848cb11c49ad85d2a93819747d`
- License: Apache-2.0
- Tests: 96 C++ unit checks + 46 SQL queries, 0 fail / 0 xfail
- Reproducibility entry: `scripts/local_check.sh`
- BENCHMARK.md is append-only and includes cells where GPU loses to CPU (cold streaming SUM in particular)

## Author note
Solo project; happy to iterate on description, build harness, or excluded_platforms list if anything in the standard CI workflow doesn't fit. The extension is built via CMake from `src/extension/duckdb_loadable.cpp` against pre-built `libduckdb-libs/`.


singhpratech:
Hi maintainers 👋 — the earlier `linux_arm64` failure (`make: *** No rule to make target 'set_duckdb_version'`) is fixed: the extension repo was missing the extension-template build harness entirely. It's now scaffolded on the C-API template (`extension-ci-tools` `c_api_extensions` makefiles, vendored C API headers, stable C_STRUCT ABI — no libduckdb link), and `ref` is pinned to the commit where that landed on `main`.

This exact workflow (DuckDB v1.5.2, ci-tools `v1.5-variegata`) is green on all four platforms on our fork: https://github.com/singhpratech/community-extensions/actions/runs/28753457602

Could you approve the workflow run when you get a chance? Thanks!

singhpratech:
👋 Friendly follow-up — this one is ready for another look.

The identical `build.yml` (DuckDB v1.5.2 / ci-tools `v1.5-variegata`) dispatched against this PR's exact branch + pinned ref on my fork is still green — freshest run today (2026-07-12): https://github.com/singhpratech/community-extensions/actions/runs/29198944838 — all four platforms (`linux_amd64`, `linux_arm64`, `osx_amd64`, `osx_arm64`) build and pass `test_release`; Windows/Wasm are correctly skipped per `excluded_platforms`.

The PR just needs a workflow approval when you have a moment. Happy to adjust anything in `description.yml`. Thanks!

mocusez:
Hi @singhpratech, thank you for this incredible work! Bringing GPU acceleration (via CUDA and Metal) to DuckDB is a highly exciting direction.

I was wondering if you have any end-to-end benchmark results using standard analytical query workloads, such as full TPC-H, SSB, or TPC-DS query sets? 

While micro-benchmarks for isolated operators (like a standalone Join or Aggregation) are helpful to demonstrate raw compute capabilities, they don't always translate directly to overall query speedups. This is especially true for architectures like Apple Silicon's UMA (Unified Memory Architecture). Under UMA, the unique trade-offs—such as zero-copy data sharing between CPU and GPU, versus shared system memory bandwidth contention—are difficult to fully manifest through single operators alone. 

A complete, end-to-end SQL query pipeline involves multiple stages (parsing, planning, data materialization, and pipeline breaks). Evaluating full queries would provide a much clearer and more realistic picture of how `gpudb` handles complex execution plans under these architectural characteristics. 

Looking forward to your thoughts and any further insights you might share!

singhpratech:
Hi @mocusez, thanks for taking a look — fair question, and it deserves a straight answer.

Short version: I don't have full TPC-H/SSB/TPC-DS suite numbers, and honestly I don't think publishing them for gpudb would mean much right now. The DuckDB C extension API doesn't expose optimizer or physical-operator hooks, so a loadable extension can't transparently swap operators inside a query plan — gpudb today accelerates explicit aggregate calls (`gpu_sum` / `gpu_min` / `gpu_max`, including GROUP BY and window contexts). If I ran the full 22-query suite, the vast majority of what's measured would just be DuckDB's own CPU pipeline. That's the classic way GPU database benchmarks have misled people and I'd rather not add to the pile — the scoping is called out explicitly in the repo README roadmap.

That said, the numbers I do publish are closer to what you're asking for than typical operator microbenches. The scorecard in [BENCHMARK.md](https://github.com/singhpratech/duckdbgpumetaldbram/blob/main/BENCHMARK.md) runs on real SF1/SF10 lineitem columns, wall clock, median of 5, against the DuckDB CLI at its actual default (16 threads on M4 Max). And since CPU and GPU share the same LPDDR5X on Apple Silicon, the bandwidth contention you mention is baked into every row. You can actually see the UMA trade-off directly in there: cold zero-copy 1B SUM roughly ties the 16-thread CPU (both contention-bound), while multi-agg fusion sustains ~475 GiB/s (~87% of peak) by amortizing one read across four aggregates. The scorecard also keeps a documented loss (50-unique GROUP BY, CPU 14× faster — the hash table sits in L1 and no GPU beats that), which I think is the only honest way to do this.

What I can add, and will: end-to-end rewritten queries through the whole DuckDB pipeline (parquet scan → filter → aggregate), Q6- and Q1-shaped with gpu_sum swapped in, with expectations set honestly — Q1's 4-group aggregate is exactly the regime where the CPU wins. I'll put those in BENCHMARK.md. Transparent full-plan acceleration is the longer-term story and realistically needs either operator hooks in the C API or a table-function path that doesn't compose well with normal SQL, which I've so far chosen not to ship.

singhpratech:
Thank you @samansmink! Really happy to see gpudb land — this one's been a labor of love, and the community-extensions infra made shipping signed multi-platform binaries genuinely painless. I'm planning an update PR soon to pick up the v0.3 improvements that shipped in the meantime. Thanks again for taking the time on the sweep. 🦆

PR #1908, opened 2026-05-14
cozip v1.1.0: bump ref to lazy ParquetReader reader

csaybar:
Summary of changes in 1.1.0:

- `read_cozip(path)` parses the cozip index at byte 0 of the archive,
  locates the `__metadata__` Parquet payload, and constructs a virtual
  path `cozip-subfile://<offset>_<size>!<underlying>`.
- A new `CozipSubFileSystem` is registered on extension Load. When
  `ParquetReader` opens the virtual path, range reads against the subfile
  are translated by adding `base_offset` and delegated to the underlying
  transport. Works over local files, https, s3, gcs, abfss, hf://.
- No temp files are written. Reads remain lazy: only the Parquet footer
  and the projected row groups are fetched.

### DIFF
--- a/extensions/cozip/description.yml
+++ b/extensions/cozip/description.yml
@@ -1,24 +1,31 @@
 extension:
   name: cozip
   description: Cloud-Optimized ZIP reader for DuckDB
-  version: 1.0.0
+  version: 1.1.0
   language: C++
   build: cmake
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1910, opened 2026-05-14
Add table_guard extension: table-level access control via allowlist

yoogoc:
## Summary

  Add the `table_guard` extension — table-level access control for DuckDB via an allowlist mechanism. Only tables explicitly added to the allowlist can be
  queried; all other table accesses are blocked with a `PermissionException`.

  ## Key Features

  - **Allowlist-based access control**: restrict queries to explicitly permitted tables only
  - **Flexible matching**: supports `catalog.schema.table`, `schema.table`, bare `table`, and `*` wildcard patterns
  - **One-time configuration**: allowlist can only be set once per connection, preventing runtime tampering
  - **Metadata filtering**: automatically filters `duckdb_tables()`, `duckdb_views()`, and `duckdb_columns()` to hide unauthorized tables
  - **Enable/disable toggle**: temporarily disable the guard without losing configuration

  ## Usage

  ```sql
  LOAD table_guard;
  PRAGMA table_guard_allow('mydb.public.patients, mydb.public.visits');

  -- Allowed
  SELECT * FROM mydb.public.patients;

  -- Blocked → PermissionException
  SELECT * FROM mydb.public.audit_log;

### DIFF
--- a/extensions/table_guard/description.yml
+++ b/extensions/table_guard/description.yml
@@ -1,0 +1,68 @@
+extension:
+  name: table_guard
+  description: Enforces table-level access control via an allowlist mechanism, blocking queries to unauthorized tables with a PermissionException
+  version: 0.1.0
+  language: C++
+  build: cmake
+  license: MIT
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads"
+  maintainers:
+    - yoogoc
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks! And welcome!

PR #1913, opened 2026-05-14
Update DICOM extension to v0.2.0

nmontesg:
This update brings the following changes to the DICOM extension:

- The extension is now built for DuckDB 1.5.2
- Add support for reading DICOM files from remote storage through the `httpfs` extension


carlopi:
Thanks!

PR #1918, opened 2026-05-17
Bump ducksmiles to 0.3.0 (morgan_fp_bits)

nkwork9999:
Adds [`morgan_fp_bits()`](https://github.com/nkwork9999/duckSMILES/blob/main/crates/smiles/src/morgan.rs) — Morgan / ECFP fingerprint as a fixed-width `BLOB`.

## Changes

- Bump version `0.2.0` → `0.3.0`
- Update ref to [`85dfe3c`](https://github.com/nkwork9999/duckSMILES/commit/85dfe3cae4ba15e9e59887a0c8d0ba16201d919e)
- Function count `37` → `38`
- Add `hello_world` examples (popcount + Tanimoto via `CAST AS BIT`)
- Document Morgan/ECFP in `extended_description`

## API

```sql
-- ECFP4 default (radius=2, 2048 bit) → 256-byte BLOB
SELECT morgan_fp_bits('CCO');

-- Custom radius / width
SELECT morgan_fp_bits('CC(=O)Oc1ccccc1C(=O)O', 3, 4096);  -- ECFP6 / 4096 bit

-- popcount (BLOB has no native bit_count, cast to BIT)
SELECT bit_count(CAST(morgan_fp_bits('CCO') AS BIT));  -- 6

-- Tanimoto similarity via BIT operators
WITH x AS (
  SELECT CAST(morgan_fp_bits('CCO') AS BIT) AS a,
         CAST(morgan_fp_bits('CCN') AS BIT) AS b
)
SELECT bit_count(a & b)::DOUBLE / bit_count(a | b) FROM x;  -- 0.3333
```

## Implementation notes

Port of RDKit's `MorganGenerator.cpp` (layered BFS + `hash_combine` + dead-atom dedup). Uses an independent hash function so bits are **not** bit-exact with RDKit, but the algorithm structure is preserved; downstream Tanimoto rankings should be comparable. Verified end-to-end via DuckDB shell against the Rust unit test popcounts.

### DIFF
--- a/extensions/ducksmiles/description.yml
+++ b/extensions/ducksmiles/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: ducksmiles
-  description: Cheminformatics toolkit for DuckDB - SMILES, InChI, MOL/SDF, PDB, SELFIES, and Wildman-Crippen LogP from SQL
-  version: 0.2.0
+  description: Cheminformatics toolkit for DuckDB - SMILES, InChI, MOL/SDF, PDB, SELFIES, Wildman-Crippen LogP, and Morgan/ECFP fingerprints from SQL
+  version: 0.3.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #1923, opened 2026-05-18
moving duckdb and libduckdb-sys from fixed 1.2.2 for 1.10502.0 (1.5.2) version)

martin-conur:
quackformers issue: https://github.com/martin-conur/quackformers/issues/22


carlopi:
Thanks!

PR #1928, opened 2026-05-19
Add quack_oauth

sipemu:
Adds the `quack_oauth` extension providing OAuth and OpenID primitives for authentication and authorization for the DuckDB quack server.

Split from #1872 per the one-extension-per-PR policy.


carlopi:
Thanks!

PR #1931, opened 2026-05-19
release: DuckDB Delta Sharing v1.5.2.8

niger-prequel:
## Summary of Changes

- [x] ensure doc in the yaml reflects previous secrets change
- [x] target correct hash which has a fix to make the credential chain deterministic


carlopi:
Thanks!

PR #1971, opened 2026-05-24
Update hive metastore extension to 0.2.0

thijs-s:
This update brings improvements to address resolution and better support for creates and inserts.


carlopi:
Thanks!

PR #2008, opened 2026-06-04
Add duck_diff extension v0.3.0

avaitla:
Adds the [duck_diff](https://github.com/avaitla/duck_diff) extension.

We have a large number of use cases where we need to see if a refactored sql gives the same output as an existing one. This could be to improve a queries performance or to move it between one database and another. Validation and diffing has been very important to solve this problem. This extension leverages duckdb's federation ability to diff two different different sources. It is built with usefulness in mind so can do safe upcasting or float rounding if needed (which happens a lot with different dataset sources).

Functions: `table_diff`, `table_diff_summary`, `schema_diff`.

- Built from the [C++ extension template](https://github.com/duckdb/extension-template), targeting DuckDB v1.5.3
- MIT licensed, no vcpkg dependencies
- Pinned ref: `fae4c5ca4a28a3074f73746e69ec203ddd489308` (main)

### DIFF
--- a/extensions/duck_diff/description.yml
+++ b/extensions/duck_diff/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: duck_diff
   description: Diff two relations off a primary key, reporting per-row and per-column changes
-  version: 0.2.0
+  version: 0.3.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


avaitla:
Curious to get thoughts and feedback, and if there are things that you see can be improved. Currently we've been using this for a few weeks internally to ensure the apis are reasonable. 

In terms of naming, some other thoughts are relation_diff or table_diff, theres also data_diff but that may sound too similar to https://github.com/datafold/data-diff

avaitla:
Wanted to see if you had any feedback on this extension @samansmink

PR #2013, opened 2026-06-05
MSSQL version to 0.2.1

VGSML:


VGSML:
@carlopi could you merge it, thanks!

PR #2034, opened 2026-06-08
Fix docs generation: well-formed CSV for extension index

oluies:
## Problem

The scheduled **Generate and Deploy Website** workflow is failing in the *Generate docs* step (e.g. [run 27123513199](https://github.com/duckdb/community-extensions/actions/runs/27123513199)) with:

```
Binder Error: Positional reference 2 out of range (total 1 columns)
##[error]Process completed with exit code 1.
```

(The `Invalid Input Error: table empty` lines in the log are the expected `CASE WHEN ... error('table empty')` guard pattern, not the failure.)

## Root cause

The final aggregation query in `scripts/generate_md.sh` reads `community_extensions.csv` by **positional column** (`#1` name, `#2` repo, `#4` description):

```sql
SELECT '['||#1||']...' as Name, '...'||#2||'...' as GitHub, #4 as Description
FROM read_csv('build/docs/community_extensions.csv');
```

That CSV was built per-extension by appending the `name,repo,ref` on one line and the description on the next, with `sed 's/$/"/'` adding a quote to **every** line. For multi-line or quote-containing descriptions (e.g. `duck_lineage`, `dicom`, `ducknng`) this emits malformed rows with comma-less continuation lines, so DuckDB's CSV sniffer collapses the file to a **single column** and the positional references blow up.

Reproduced locally by building the full CSV from all `extensions/*/description.yml` and running the final query → identical `total 1 columns` error.

## Fix

Build each CSV row in a single `printf`: flatten the description to one line and escape embedded double quotes as `""` (RFC 4180). With the fix, all rows parse and the final query succeeds. Also added a comment documenting the positional-column invariant at the final query.

No other logic (alias handling, per-extension docs) is touched.

### DIFF
--- a/scripts/generate_md.sh
+++ b/scripts/generate_md.sh
@@ -109,9 +109,15 @@ do
          if [[ "${REPOSITORY}" == "duckdb/duckdb-extension-alias" ]]; then
             extension=$(cat extensions/$extension/description.yml | yq -r ".repo.canonical_name")
          fi
-         cat extensions/$extension/description.yml | yq ".extension.name, .repo.github, .repo.ref" | xargs printf '%s,%s,%s,"' >> $EXTENSIONS_CSV
-
-         cat extensions/$extension/description.yml | yq -r '.extension.description' | sed 's/$/"/'  >> $EXTENSIONS_CSV
+         # Build the CSV row in one shot. The description is flattened to a single
+         # line and embedded double quotes are escaped ("" per RFC 4180), otherwise
+         # multi-line or quote-containing descriptions emit malformed rows that make
+         # DuckDB's CSV sniffer detect a single column and break the final query below.
+         CSV_NAME=$(yq -r '.extension.name' extensions/$extension/description.yml)
+         CSV_REPO=$(yq -r '.repo.github' extensions/$extension/description.yml)
+         CSV_REF=$(yq -r '.repo.ref' extensions/$extension/description.yml)
+         CSV_DESC=$(yq -r '.extension.description' extensions/$extension/description.yml | tr '\n' ' ' | sed 's/  */ /g; s/ *$//; s/"/""/g')
+         printf '%s,%s,%s,"%s"\n' "$CSV_NAME" "$CSV_REPO" "$CSV_REF" "$CSV_DESC" >> $EXTENSIONS_CSV
          echo "" >> $EXTENSION_README
          cat extensions/$extension/description.yml >> $EXTENSION_README
          echo "" >> $EXTENSION_README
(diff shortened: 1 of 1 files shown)


oluies:
## Security review

Scope: the single change to `scripts/generate_md.sh` in this PR (CI docs-generation script; inputs come from reviewed/merged `extensions/*/description.yml`).

**No security vulnerabilities introduced — net slightly positive.**

- **Command injection** — not introduced. New code uses `printf '%s' "$VAR"` with quoted command-substitution variables, which is *safer* than the old `yq | xargs printf` (xargs interprets quotes/backslashes and word-splits untrusted text). Attack surface decreased.
- **SQL injection** — not present. The final DuckDB query reads the CSV via `read_csv(...)` and references columns positionally (`#1`/`#2`/`#4`) as data, never concatenated into SQL text. Unchanged by this diff.
- **CSV/formula injection & stored-XSS into the generated page** — pre-existing concerns tied to merged `description.yml` content; not introduced or worsened. The new RFC-4180 quote-doubling (`s/"/""/g`) and newline-flattening marginally *reduce* output-injection and CSV column-desync risk.
- **Path traversal** — no path construction is touched by this diff.

No findings meet the reporting threshold.

## Also fixes #1677

The same malformed-CSV root cause explains #1677 (the published `community_extensions.csv` showing far fewer extensions than the docs page). Multi-line / quote-containing descriptions (first break at `dicom`'s `"...Medicine)."`) produce unterminated quoted fields, so a strict CSV parser slurps subsequent rows into one field and truncates the list. With this fix the regenerated CSV has **0 malformed rows** and all 252 extensions parse cleanly.

Closes #1677

carlopi (approved):
Thanks! I think this makes sense.

oluies:
**Correction:** I earlier wrote that this PR closes #1677 — that is **not** accurate, and I've removed `Closes #1677` from the description.

This PR fixes the `generate_docs` crash (`Binder Error: ... total 1 columns`), which unblocks the markdown docs pipeline. But #1677 (the stale published `community_extensions.csv` at duckdb.org) is an independent **deploy-side** problem: that CSV is a committed static file in `duckdb/duckdb-web` (`community_extensions/community_extensions.csv`) and no active workflow regenerates it. The active deploy (`duckdb-web/.github/workflows/community_extension_docs.yml`) only PRs the markdown list/pages; the only `aws s3 cp` of the CSV lives in `community-extensions/deploy_docs.yml` gated `if: github.repository == 'duckdb/duckdb-web'`, which is unreachable. A separate fix in duckdb-web is needed for #1677.

PR #2041, opened 2026-06-09
Add dbn extension (Databento DBN file reader)

tbeason:
Adds the **dbn** extension: https://github.com/tbeason/duckdb-dbn

`dbn` reads [Databento](https://databento.com) DBN (Databento Binary Encoding) market-data files — including the Zstd-compressed `.dbn.zst` captures Databento ships — directly as SQL tables, with no conversion step. It supports every DBN market-data schema (trades, mbo, mbp-1/mbp-10, tbbo, bbo/cbbo/cmbp-1/tcbbo, ohlcv, status, imbalance, statistics, definition), DBN versions 1–3, glob/multi-file reads, filter & projection pushdown, and inline symbol resolution for live captures. DBN "undefined" sentinels are surfaced as SQL `NULL`, and decoding is differentially verified bit-identical to the official `databento` Python decoder across every schema and version.

- Built against DuckDB v1.5.3; the extension's own CI (DuckDB's reusable distribution workflow) is green on the pinned commit across Linux x64/arm64, Windows (MSVC + MinGW), and macOS x64/arm64.
- `ref` is pinned to the commit of the `v0.1.0` tag.
- No vcpkg dependencies.
- Wasm platforms are excluded: the reader currently uses stdlib file I/O, which has no meaningful filesystem in the browser. (It does compile for Wasm; the exclusion can be lifted if/when the reader moves to DuckDB's FileSystem API.)
- MIT licensed. This is an independent reader, not affiliated with Databento.

### DIFF
--- a/extensions/dbn/description.yml
+++ b/extensions/dbn/description.yml
@@ -1,0 +1,55 @@
+extension:
+  name: dbn
+  description: Read Databento DBN (Databento Binary Encoding) market-data files, including Zstd-compressed .dbn.zst captures, directly as SQL tables
+  version: 0.1.0
+  language: C++
+  build: cmake
+  license: MIT
+  # The readers open files with the C++ stdlib filesystem, which does not
+  # exist in the browser — exclude Wasm rather than ship a non-functional
+  # binary (it does compile, so this can be revisited if the reader moves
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2045, opened 2026-06-10
Update bvh2sql source ref

nkwork9999:
## Summary

Update `bvh2sql` source ref:

- `bvh2sql`: `b72eec2` -> `1253d33`

`pic2vec` is split into a separate PR because the community-extensions build script only accepts one changed descriptor per PR. `duckorch` is intentionally handled separately in #2049.


carlopi:
Thanks!

PR #2046, opened 2026-06-10
Fix: Simplify dicom extension documentation

nmontesg:
The documentation for the `dicom` extension was not rendering correctly in the DuckDB CE website. This update fixes issues with some documentation fields and simplifies it (points the user to the repo’s `README` for further details).


carlopi:
Thanks!

PR #2047, opened 2026-06-10
H5db, minor wording change to rerun deploy workflow

jokasimr:
As title says, this PR is only to let the deploy workflow rerun.
 
In the last run the workflow failed in the `build docker image` step: https://github.com/duckdb/community-extensions/actions/runs/27193945750/job/80285265282 with error `Failed to download metadata for repo 'epel'`.

I don't know what caused the previous failure, but because it occurred so early in the pipeline I assume it did not have to do with the extension itself.


carlopi:
Thanks!

PR #2048, opened 2026-06-11
zim: bump to v0.4.0

teaguesterling:
Bumps the `zim` extension from v0.2.0 to **v0.4.0**.

Since v0.2.0 the extension gained:

- **Parallel scans** and **remote archives** — read `.zim` files straight from S3/HTTP via byte-range requests (only the bytes a query touches are fetched), in addition to local files (v0.3.0).
- **SQL filter pushdown** — `WHERE path`/`title`/`mimetype` predicates are pushed into libzim (v0.4.0, #7 upstream).
- **Warm remote-handle caching with a clean process shutdown** — recently-used remote handles stay warm between calls; the handle pool is scoped to the DuckDB database so it releases before the FileSystem is torn down (v0.4.0, #8 upstream).

Descriptor changes only: `version` → 0.4.0, `ref` → the v0.4.0 release commit. `vcpkg_commit` is unchanged (`84bab45d`, the `extension-ci-tools@v1.5-variegata` default the extension's own CI builds against). Linux + macOS + Wasm all pass on the upstream repo's release pipeline; Windows remains excluded (libzim/icu MSVC porting is still open).

### DIFF
--- a/extensions/zim/description.yml
+++ b/extensions/zim/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: zim
-  description: Read .zim (Kiwix / openZIM) archives directly in DuckDB via libzim — offline Wikipedia, WikiMed, Stack Exchange, iFixit, and more, with a zim:// filesystem and full-text search.
-  version: 0.2.0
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2051, opened 2026-06-12
Add duck_diff extension (v0.3.0)

avaitla:
Adds the [duck_diff](https://github.com/avaitla/duck_diff) extension.

`duck_diff` diffs two relations (tables, queries) off a primary key. For each key it reports whether the row is identical, different, or exists only on one side, with both a JSON `diff_data` summary of the changed columns and expanded per-column `<c>_left` / `<c>_right` / `<c>_diff_status` columns. Includes `table_diff`, `table_diff_summary`, and `schema_diff` functions, with options for numeric tolerance, timestamp precision, column subsets, and composite keys.

- Language: C++ (extension template)
- Build: cmake
- License: MIT
- Pinned ref: `3f278772095f7abb1c67a7cd17e039cd48dee20c` (v0.3.0)

### DIFF
--- a/extensions/duck_diff/description.yml
+++ b/extensions/duck_diff/description.yml
@@ -1,0 +1,75 @@
+extension:
+  name: duck_diff
+  description: Diff two relations (tables, queries) off a primary key, reporting per-key status (identical/different/left_only/right_only), a JSON summary of changed columns, and per-column left/right values.
+  version: 0.3.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - avaitla
+
+repo:
+  github: avaitla/duck_diff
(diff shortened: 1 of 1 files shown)


avaitla:
The Windows (MSVC) CI failure here is not specific to this extension — it's an infra-wide breakage affecting all current Windows PRs. DuckDB v1.5.3's vendored `fmt` uses `stdext::checked_array_iterator`, which MSVC 14.51 (now default on `windows-latest`) removed. The fix is duckdb/duckdb#22706 but it's on `main`, not in the pinned v1.5.3 tag.

Tracked in #2061 (with evidence and suggested fixes). Nothing to change in this PR.

carlopi:
Thanks!

PR #2053, opened 2026-06-12
Update pic2vec source ref

nkwork9999:
Update pic2vec source ref
- pic2vec: 9705994 -> 81c15d5
- Split from #2045 because the community-extensions build workflow accepts one changed descriptor per PR.


carlopi:
Thanks!

PR #2074, opened 2026-06-15
add concurrency rule

c-herrewijn:
implements: https://github.com/duckdblabs/duckdb-internal/issues/7036

in-progress workflow runs are cancelled if additional commits are added to a PR


carlopi (approved):
Thanks!

Let's wait for v1.5.4 to go out, then CI will pass, and this will not have an effect on release iterations.

carlopi:
Thanks!

PR #2075, opened 2026-06-16
Add duckton extension

arghaffari:
## duckton

**Can you trust a query result you didn't compute yourself?** `duckton` turns DuckDB into a node of a **trustless peer-to-peer query grid** that answers that with cryptography instead of faith.

A query is dispatched to several independent hosts that run it redundantly over **QUIC** (TLS 1.3, mutually authenticated). Each result is reduced to a canonical, **order-independent BLAKE3 hash**, and the answer is accepted only once a configurable **quorum** of those hashes agree byte-for-byte — so a single broken or actively dishonest host can't slip a wrong result past the quorum. It's distributed compute you can actually *verify*, all from plain SQL table functions, with no external daemon to run.

### Why it matters
- **Trustless results.** Redundant execution + quorum agreement on canonical result hashes means correctness doesn't depend on trusting any one host.
- **Zero-config and safe by default.** With no peers configured, `p2p_query` runs on a **free, in-process locked-down DuckDB** — no network egress, no local-filesystem access, configuration locked so an untrusted query can't pry it back open. Useful and safe the moment you `LOAD` it.
- **Observable verification.** Hedged execution races `k` workers, accepts the first to reach quorum, and `RESET`s the losers; `p2p_query_meta()` exposes whether it ran locally, whether quorum verified it, the winning host, the agreed hash, and participant/receipt counts.
- **Pay for compute only if you want to.** Public jobs are free and off-chain; paid compute is **opt-in and default-off**.

### Example
```sql
INSTALL duckton FROM community;
LOAD duckton;

SELECT * FROM p2p_info();                         -- protocol identity
SELECT * FROM p2p_query('SELECT 42 AS answer');   -- runs locally, verified
```

### Trust & (optional) economics
Hosts earn a confidence-aware (Wilson-shrunk) reputation from **signed, gossiped receipts**; selection weighs reputation, attestation tier (L0/L1/L2), vouchers and stake, and each node keeps its own local deny-list (`p2p_block` / `p2p_unblock`) — there is no central authority in the data path.

When you want guaranteed, accountable compute, jobs can be **paid** through a per-job escrow on **The Open Network (TON)**. The economics are enforced on-chain: the escrow must cover all parties, with the platform fee (**15%**) and the non-winner verifier commission (**5%** per agreeing replica) paid out correctly on settlement, plus stake/slashing and per-epoch Merkle anchoring of job records. The free grid never instantiates a chain client.

### Details
- **Repo:** `Angelerator/duckton` @ `062dc7690cdf44675787186127e0bd2af42816c4` (tag `v0.2.1`)
- **Language:** Rust · **Build:** cmake (templated with `extension-ci-tools`) · **License:** Apache-2.0
- **API:** built on DuckDB's **stable C extension API** (loadable extension)
- **Toolchains:** `rust;python3`
- **Excluded platforms:** wasm (no async sockets), musl, and MinGW/RTools Windows — the QUIC/TLS stack (quinn + rustls + ring) + async runtime aren't supported there. Native glibc Linux, macOS, and MSVC Windows are targeted.

Built and tested locally against DuckDB v1.5.4 (`make configure && make release`, LOAD + SQLLogicTest pass on osx_arm64; cross-built osx_amd64).

### DIFF
--- a/extensions/duckton/description.yml
+++ b/extensions/duckton/description.yml
@@ -1,0 +1,101 @@
+extension:
+  name: duckton
+  description: A peer-to-peer DuckDB query grid — dispatch SQL to a swarm of hosts that run it redundantly and must reach a verifiable quorum, with optional TON-settled paid execution
+  version: 0.2.1
+  language: Rust
+  build: cmake
+  license: Apache-2.0
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;linux_amd64_musl;linux_arm64_musl;windows_amd64_mingw;windows_amd64_rtools"
+  requires_toolchains: "rust;python3"
+  maintainers:
+    - Angelerator
+
+repo:
+  github: Angelerator/duckdb-p2p
+  ref: 062dc7690cdf44675787186127e0bd2af42816c4
+
+docs:
+  hello_world: |
+    -- Inspect the protocol identity exposed by the extension.
+    SELECT * FROM p2p_info();
+    ┌───────────────────────┬──────────────┐
+    │          key          │    value     │
+    │        varchar        │   varchar    │
+    ├───────────────────────┼──────────────┤
+    │ protocol_name         │ duckdb-p2p   │
+    │ protocol_version      │ 1.0.0        │
+    │ min_supported_version │ 1.0.0        │
+    │ schema_version        │ 1            │
+    │ extension_version     │ 0.2.1        │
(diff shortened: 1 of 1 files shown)


carlopi:
I need to close and reopen the PR to have this retested against proper v1.5.4 commit.

carlopi:
Thanks!

PR #2079, opened 2026-06-18
Add laterite_ags4 extension

niko86:
laterite_ags4 — AGS4 ground-investigation files as typed, join-ready SQL tables, straight from DuckDB.

AGS4 is how the UK geotechnical industry exchanges borehole and lab data: a grouped, comma-delimited text format where every value is a quoted string and the links between groups (locations, samples, lab tests, geology) live in repeated key columns. Getting it into a database usually means writing a parser, casting dozens of columns by hand, and working out which keys join to what. This extension does that part for you.

read_ags('site.ags', 'LOCA') returns a DuckDB table whose columns are already typed from the AGS data dictionary — ground levels as DOUBLE, dates as TIMESTAMP — and every row carries an _id and _parent_id. Those ids are a hash of the AGS key columns, not something allocated at runtime, so a sample from the SAMP group and its borehole from LOCA land on the same id without you naming the join columns. That holds whether you read the groups in one query or several, and from one file or many.

What you get:
- Columns typed from the dictionary instead of strings.
- Keys that join by construction: s._parent_id = l._id works for any parent/child pair in the AGS schema — no ON clause to reverse-engineer.
- Remote files through DuckDB's filesystem — s3:// and http(s):// once you LOAD httpfs.
- The dictionary and relationship graph as queryable tables (ags_dictionary, ags_relationships), plus per-file metadata (ags_groups, ags_headings).
- Validation when you ask for it (ags_validate), out of the way when you don't.
- WebAssembly support via read_ags_text(content, group) — same born-typed output, but the AGS text is passed as an argument (a bound parameter in a browser app, which already holds the file bytes), so it needs no filesystem.

Example (native):
```
INSTALL laterite_ags4 FROM community;
LOAD laterite_ags4;

SELECT loca_id, loca_gl FROM read_ags('site.ags', 'LOCA') WHERE loca_gl > 50.0;

SELECT l.loca_id, s.samp_top, s.samp_ref
FROM read_ags('site.ags', 'SAMP') s
JOIN read_ags('site.ags', 'LOCA') l ON s._parent_id = l._id;
```

In the browser (DuckDB-WASM), the app supplies the AGS text:
```
SELECT loca_id, loca_gl FROM read_ags_text($content, 'LOCA') WHERE loca_gl > 50.0;
```

Details:
- Repo: niko86/laterite-duckdb @ a367ecec7bf0cab90ad63fa341a4234e3bd5af19 (tag v0.4.0)
- Language: Rust · Build: cargo (templated with extension-ci-tools) · License: MIT
- API: DuckDB C extension API (loadable extension). Native builds use the unstable C API for filesystem access (path + remote readers); the WASM build is a stable-API subset — read_ags_text + ags_dictionary + ags_relationships — that's forward-compatible. The path/remote readers return on WASM once its engine reaches native's.
- Toolchains: rust
- Excluded platforms: linux_amd64_musl (static-musl, untested). Native glibc Linux, macOS, MSVC Windows, and DuckDB-WASM are targeted.
- Built and tested against DuckDB v1.5.4 — native (make configure && make release && make test, full sqllogictest suite incl. read_ags_text, on osx_arm64) — and DuckDB-WASM 1.5.1, where read_ags_text returns correct typed rows in the browser.

### DIFF
--- a/extensions/laterite_ags4/description.yml
+++ b/extensions/laterite_ags4/description.yml
@@ -1,0 +1,65 @@
+extension:
+  name: laterite_ags4
+  description: Read AGS4 geotechnical data files as typed, UUID-keyed tables directly from SQL — born-typed columns, deterministic content-addressed keys that join across groups by construction, embedded AGS dictionary, and opt-in validation. Local, http(s):// and s3:// (with httpfs).
+  version: 0.4.0
+  language: Rust
+  build: cargo
+  license: MIT
+  # wasm + static-musl are untested for this extension's bundled-engine-free
+  # loadable build; native platforms first, wasm is a follow-up.
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;linux_amd64_musl"
+  requires_toolchains: "rust"
+  maintainers:
+    - niko86
+
+repo:
+  github: niko86/laterite-duckdb
+  # Pin to the release commit SHA once the repo is pushed + tagged (community-
+  # extensions pins a SHA, not a branch). Placeholder until then:
+  ref: 7875a8507029c02677436960962b482080c46301
+
+docs:
+  hello_world: |
+    -- Read a group as a typed table (columns typed from the file's own TYPE row):
+    SELECT loca_id, loca_gl
+    FROM read_ags('site.ags', 'LOCA')
+    WHERE loca_gl > 50.0;
+
+    -- Join across groups on the deterministic keys — no shared state, joins by
+    -- construction (every SAMP row's _parent_id equals its LOCA's _id):
+    SELECT l.loca_id, s.samp_ref, s.samp_top
+    FROM read_ags('site.ags', 'SAMP') s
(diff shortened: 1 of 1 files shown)


carlopi:
I think the wasm build has the same issues as `polyglot`, see https://github.com/duckdb/extension-ci-tools/issues/385. We'll need to have a look, in the meantime you could consider skipping the `wasm_*` platforms or wait for this to be fixed.

niko86:
> I think the wasm build has the same issues as `polyglot`, see [duckdb/extension-ci-tools#385](https://github.com/duckdb/extension-ci-tools/issues/385). We'll need to have a look, in the meantime you could consider skipping the `wasm_*` platforms or wait for this to be fixed.

I removed WASM for now, hopefully now everything else is in order.

carlopi:
Thanks! And welcome!

niko86:
Thank you 😊 DuckDB forms a core part of my python library so it made sense to make an extension.

PR #2087, opened 2026-06-19
duck_hunt: bump to 1.10.0

teaguesterling:
Bumps `duck_hunt` from 1.9.0 to 1.10.0.

Both `ref` and `andium` are updated to `22c142f` (tag `v1.10.0`).

What's in this release:
- duckdb v1.5.3 support (bumped from v1.4.4).
- duckdb-main build compatibility (templated `Catalog::GetEntry<T>`, `Identifier`-wrapped lookups gated by `__has_include`).
- A WASM symbol-resolution + live-load test harness under `test/wasm/` (repo-only; does not change the shipped extension).

No new parsers or user-facing functional changes. Verified to build against both v1.5.x and the v1.4 "Andium" LTS line (templated `GetEntry<T>` and `lambda` syntax both exist in v1.4).

### DIFF
--- a/extensions/duck_hunt/description.yml
+++ b/extensions/duck_hunt/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: duck_hunt
   description: Parse and analyze test results, build outputs, and CI/CD pipeline logs from 110+ formats with severity filtering, format auto-detection, and context extraction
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2091, opened 2026-06-20
zim: bump to v0.4.1

teaguesterling:
Bumps the zim extension to v0.4.1.

Fixes the DuckDB-WASM build, which installed but failed to load because libzim (and its transitive zstd/lzma) were not linked into the `-sSIDE_MODULE=2` module. `CMakeLists.txt` now resolves libzim's static link set from pkg-config and feeds it to the WASM link (grouped for ICU's circular refs); a runtime-free static symbol gate is wired into CI. ICU dead-strips (libzim's read path doesn't use it), so the `.wasm` stays ~0.5 MB. No functional/native changes.

Note: this makes the extension *load* under duckdb-wasm; reading a ZIM in the browser is separate (libzim's local mmap path; tracked for a later release).

Release: https://github.com/teaguesterling/duckdb_zim/releases/tag/v0.4.1


carlopi:
Thanks!

PR #2090, opened 2026-06-20
duckton: bump to v0.6.0

arghaffari:
Bumps the published **duckton** extension from v0.5.2 to **v0.6.0**.

Updates `extensions/duckton/description.yml`:
- `repo.ref` → `c0c6087f6e5267c4c06d761f088504181ebba99f` (tag `v0.6.0` on `Angelerator/duckton`)
- `version` → `0.6.0` (and the `extension_version` shown in the `hello_world` example)

### What's new in v0.6.0
- **Smart size-aware routing with robust failover** — jobs are placed by result size and re-dispatched around slow or failing hosts.
- **Process-wide capacity governor** — caps concurrent compute across the node.
- **Presigned-URL credential mode** — optional, for object-store access.
- **Stronger stake weighting** — stake counts for more in host selection.
- **Web/console refresh** and a **fresh on-chain proof**.

Built and tested against **DuckDB v1.5.4** (the registry's current version).

PR #2092, opened 2026-06-20
duckton: bump to v0.6.1 (fix two-node grid hang)

arghaffari:
Bumps the `duckton` community extension from **v0.6.0** to **v0.6.1**, a bug-fix release.

- `repo.ref` → `040c45638927fb56053033409bb28bc3796f4a1b` (tag `v0.6.1` on `Angelerator/duckton`)
- `version` / `extension_version` → `0.6.1`

### What v0.6.1 fixes
v0.6.0 (currently in the registry, #2090) contains a **two-node-grid hang**: a requester could spin forever re-dispatching to a dual-role node that rejected a full-budget job. v0.6.1 fixes this with a **bounded no-progress guard** plus a **governor served-ceiling clamp**, so dispatch always makes forward progress or fails cleanly.

### Also in v0.6.1: secure spill
- `max_temp_directory_size` cap on on-disk spill
- `temp_file_encryption` (spill encrypted at rest)
- temp-dir / sandbox alignment so spill stays inside the locked-down sandbox

### Build
Built and tested against **DuckDB v1.5.4**. The diff is limited to `extensions/duckton/description.yml`.

### DIFF
--- a/extensions/duckton/description.yml
+++ b/extensions/duckton/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: duckton
   description: Turn DuckDB into a node of a trustless peer-to-peer query grid — SQL is cross-checked by independent hosts to a verifiable quorum over QUIC, with optional pay-per-query compute settled on TON
-  version: 0.6.0
+  version: 0.6.1
   language: Rust
   build: cmake
   license: Apache-2.0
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2096, opened 2026-06-20
scalarfs: bump to 1.4.1

teaguesterling:
Bumps `scalarfs` from 1.4.0 to 1.4.1 (patch). `ref`, `andium`, and `ref_next` all updated to tag `v1.4.1`.

What's in this release (compatibility + test portability; no new features):
- duckdb v1.5.3 build support (stable build bumped from v1.4.4).
- duckdb-main build compatibility, guarded by `__has_include` so v1.4/v1.5 builds are unaffected (handles duckdb main's typed `Identifier` for `user_variables` keys, `child_list_t` STRUCT-field keys, and the COPY bind path).
- Test portability: a missing-file error assertion now matches the common errno substring "no such file or directory" instead of a backend-specific prefix, so it passes on both native and wasm.

Builds clean on Linux/macOS/wasm for v1.5.3 and on the v1.4 LTS line. (The duckdb-main canary is paused upstream while duckdb's Identifier migration is in flight.)

### DIFF
--- a/extensions/scalarfs/description.yml
+++ b/extensions/scalarfs/description.yml
@@ -5,6 +5,9 @@ extension:
   language: C++
   build: cmake
   license: MIT
+  # Windows excluded: duckdb's vendored fmt fails on the build runner's MSVC
+  # (stdext checked_array_iterator removed) -- upstream, not scalarfs.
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2100, opened 2026-06-21
Update DICOM extension to DuckDB v1.5.4

nmontesg:
This update includes:

* Updating DuckDB to v1.5.4
* Introduction of a `dicom` secret type to store credentials to communicate with other DICOM modalities (e.g. PACS systems).
* Introduction of a `query_dicom` table function to retrieve DICOM datasets from other DICOM modalities using C-FIND commands.


carlopi:
Thanks!

PR #2118, opened 2026-06-22
qvd: update to 1.0.2

snouhaud:
Update the `qvd` extension to **1.0.2**.

Changes since 1.0.0 (read-path performance, no API change, same DuckDB v1.5.4 target):
- Streaming **dictionary** decode: convert each symbol once instead of per row.
- Streaming **projected index** decode: extract only the projected columns' fields per record (16–30× faster on narrow queries over large files), bounded memory unchanged.

`repo.ref` points to tag `v1.0.2` (commit `4f5cc8f`). Only `extensions/qvd/description.yml` changes (`version` 1.0.0 → 1.0.2, `ref`, and a refreshed perf line).

### DIFF
--- a/extensions/qvd/description.yml
+++ b/extensions/qvd/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: qvd
   description: Read (streaming) and write Qlik QVD files
-  version: 1.0.0
+  version: 1.0.2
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2123, opened 2026-06-23
Bump rusty_quack to v1.5.4

mlafeldt:


extensions/rusty_quack/description.yml
```diff
@@ -14,7 +14,7 @@ extension:
 repo:
   github: duckdb/extension-template-rs
   andium: 2381c7e1c5026f21c92b7ffe3ae7a6b1d8d177c0
```
mlafeldt, line 16: This still points at 1.4.3. The only way to bump it to 1.4.5 would be to create (and _maintain_) yet another `v1.4-andium` branch. Worth it? @carlopi

carlopi: I think it's OK like this, thanks!

carlopi:
Thanks!

PR #2127, opened 2026-06-23
Update laterite_ags4 to 0.5.0

niko86:
Updates laterite_ags4 to 0.5.0.

validate_ags now takes `warnings := true` / `fyi := true` to include the WARNING and FYI severity tiers; it defaults to errors only. The `edition` named parameter is renamed to `dict_version` on validate_ags and certify_ags so the SQL surface matches the rest of the AGS toolchain.

Pinned to niko86/laterite-duckdb@3527538 (tag v0.5.0); wasm and linux_amd64_musl stay excluded as before.


carlopi:
Thanks!

PR #2128, opened 2026-06-23
finetype: republish at v0.6.36 (244 types, accurate ft_ docs)

hughcameron:
Bumps the published `finetype` extension from **v0.6.23 to v0.6.36** (`ref: v0.6.36`, 244 types across 7 domains) and corrects the docs to match real `ft_` output, verified against the current `finetype-duckdb` source.

### Doc corrections
- `ft_profile`'s output column is `type`, not `detected_type`
- nested `STRUCT`/`LIST` columns return a flatten hint in the `type` column (with `confidence` / `duckdb_type` `NULL`), ordered by `column_name` — not a blank row
- `ft_validate` JSON Schema must be keyed under a top-level `properties` object; the inline examples now include it
- `ft_validate_text` returns `STRUCT(valid, constraint, message)`, not `VARCHAR`
- alias note no longer claims a single-release window (the un-prefixed `finetype*` scalars remain registered)

Single-file change to `extensions/finetype/description.yml`. The `v0.6.36` tag is published on `meridian-online/finetype`, so CI can build from it.

### DIFF
--- a/extensions/finetype/description.yml
+++ b/extensions/finetype/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: finetype
-  description: Semantic type classification — detects 240 data types (emails, URLs, dates, UUIDs, currencies, etc.) from raw strings
-  version: 0.6.23
+  description: Semantic type classification — detects 244 data types (emails, URLs, dates, UUIDs, currencies, etc.) from raw strings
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2133, opened 2026-06-25
Add `pdf` extension

asubbarao:
Adds the **`pdf`** community extension: read and extract content from PDF files directly in SQL, built on **Poppler** (parsing/rendering/metadata) and **Tesseract + Leptonica** (OCR of scanned PDFs).

**Repo:** https://github.com/asubbarao/duckdb-pdf

### Functions
- Table functions: `read_pdf` (page text + dimensions), `read_pdf_lines` (layout-preserving lines), `read_pdf_meta` (document metadata), `read_pdf_words` (word bounding boxes + fonts), `read_pdf_tables` (precision-first table extraction on digital PDFs)
- Scalar functions: `pdf_to_text` (with optional `layout`), `pdf_to_html`, `pdf_to_xml`, `pdf_to_svg`
- All table functions accept a single path, a list, or a glob; shared named params for page range, password, layout, and OCR knobs.

### Platforms
Builds and tests pass on **Linux (x86_64, arm64), macOS (x86_64, arm64), and Windows x64 (MSVC)** — all dependencies resolved via vcpkg and statically linked. mingw/rtools, windows_arm64, and wasm are excluded (toolchain/linkage constraints).

### OCR note
OCR requires a Tesseract language model at runtime (located via `TESSDATA_PREFIX`); package managers don't ship one, so the extension raises a clear, actionable error when a model is absent rather than returning empty text. The repo's OCR tests are gated on a model being present.

### License
GPL-2.0-or-later (statically links Poppler, which is GPL-2.0).

### DIFF
--- a/extensions/pdf/description.yml
+++ b/extensions/pdf/description.yml
@@ -26,7 +26,7 @@ extension:
 
 repo:
   github: asubbarao/duckdb-pdf
-  ref: 323af182abef7df45a886e88e9a67a491b61db83
+  ref: 6e7dbcc12f77d2670e1bab32462bc684afc44932
 
 docs:
   hello_world: |
--- a/extensions/pdf/description.yml
+++ b/extensions/pdf/description.yml
@@ -1,13 +1,21 @@
+# Community-extensions submission descriptor for the `pdf` extension.
+#
+# To submit this extension to the DuckDB community-extensions registry:
+#   1. Fork https://github.com/duckdb/community-extensions
+#   2. Copy this file to extensions/pdf/description.yml in your fork
+#   3. Set repo.ref below to the exact git SHA of the release commit in
+#      asubbarao/duckdb-pdf that the registry should build from
+#   4. Open a pull request against duckdb/community-extensions
+
 extension:
   name: pdf
   description: >-
     Read and extract content from PDF files. Provides table functions for
-    page-level text (read_pdf), layout-preserving lines (read_pdf_lines),
-    document metadata (read_pdf_meta), word-level bounding-box extraction
-    (read_pdf_words), and table detection (read_pdf_tables), plus scalar
-    functions for full-document conversion to plain text, HTML, XML, and SVG
-    (pdf_to_text, pdf_to_html, pdf_to_xml, pdf_to_svg). Supports OCR of scanned
-    PDFs via Tesseract.
+    page-level text (read_pdf), document metadata (read_pdf_meta), word-level
+    bounding-box extraction (read_pdf_words), and table detection
+    (read_pdf_tables), plus scalar functions for full-document conversion to
+    plain text, HTML, XML, and SVG (pdf_to_text, pdf_to_html, pdf_to_xml,
+    pdf_to_svg). Supports OCR of scanned PDFs via Tesseract.
   version: 0.1.0
   language: C++
   build: cmake
@@ -24,7 +32,7 @@ extension:
 
 repo:
   github: asubbarao/duckdb-pdf
-  ref: c0956afb5e46e1013a6c105aa67b02b3ef3b053c
+  ref: 3ffb8d85c4ea4ce65876657027cd49719e7d8808
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


asubbarao:
Hi @carlopi 👋 — this is my first community-extensions PR (new `pdf` extension: Poppler text/lines/words/tables + Tesseract OCR). It's stuck on `action_required`, so CI can't run yet — would you mind approving the workflow? I just pushed a fix for the gperf/fontconfig vcpkg build failure that the first run surfaced (poppler now builds with `default-features: false` + `zlib`, dropping the fontconfig → gperf autoreconf dependency). Happy to address anything in review. Thanks!

asubbarao:
Update: pushed two build fixes since the first run, ref now at `ad440c5`. (1) poppler built with `default-features: false` + `zlib` to drop the fontconfig→gperf autoreconf failure; (2) link tesseract's curl + libarchive transitive deps via pkg-config's static closure (`--start-group` on Linux) — the prior run got past the dep build and only failed at the final link. Also hardened input validation. Would you mind approving the workflow run once more? Thanks for the patience 🙏

asubbarao:
Pushed a fix for the Linux link failure (ref now `5bfd93e`).

**Root cause:** vcpkg builds tesseract with CMake, and the generated `tesseract.pc` doesn't declare the full transitive static closure — `libtesseract.a` references libcurl + libarchive (and those pull in ssl/crypto/zstd/lzma/…), but `pkg-config --static` omitted them. The loadable `.so` linked fine (shared libs tolerate undefined symbols); the `duckdb`/`unittest`/`plan_serializer` executables did not.

**Fix:** stop trusting the `.pc` link line — glob every archive vcpkg installed and link the whole set (in a `--start-group` on Linux), attached to the static extension target so the closure propagates into the executables. Over-linking is harmless for a static link. This mirrors the existing Windows-branch strategy. Verified the system-path build + full test suite locally.

Would appreciate a workflow approval when you get a chance 🙏

asubbarao:
Update (ref `323af18`): the Linux **link** failure is fixed (link the full vcpkg static archive set). That surfaced a deeper issue the x64 container caught — `read_pdf_tables` and `pdf_to_text` shelled out to the `pdftotext`/`pdftohtml`/`pdftocairo` binaries, which aren't installed in CI or on most users' machines.

**Made the extension fully self-contained** — everything now runs on the linked poppler-cpp library, no runtime process dependency:
- `read_pdf_tables` reconstructs tables from `page::text_list()` positioned words (same precision-first heuristic).
- `pdf_to_text` uses `page::text()` directly.
- Dropped `pdf_to_html`/`pdf_to_xml`/`pdf_to_svg` (external-binary only, no poppler-cpp equivalent, not part of the value prop).
- Deleted the posix_spawn CLI bridge entirely.

Full suite passes locally (87 assertions). Would appreciate a workflow approval when you have a moment 🙏

asubbarao:
Update (ref `6e7dbcc`): **both Linux arches now build green** 🎉 — the static-link + self-contained refactor cleared x64/arm64 Linux. Two platform links surfaced once Linux passed, both now fixed:

- **macOS**: poppler-cpp's `ustring::to_utf8` references the system `libiconv` (`_iconv*`), which isn't a vcpkg archive — added `-liconv`.
- **Windows**: the `x64-windows-static-release` triplet uses the static CRT (/MT), but the extension objects were pinned to /MD → `LNK2005` duplicate-CRT symbols. Now the CRT is selected from the triplet (static for `-static-release`).

Apologies for the staged fixes — I don't have macOS/Windows vcpkg-static runners locally, so CI is surfacing these one platform at a time. Would appreciate another workflow approval when you can 🙏

asubbarao:
Update (ref `03380d1`): `-liconv` fixed the macOS `.so` link (confirmed — iconv no longer undefined), which let the executable links proceed and surface libcurl's macOS framework deps (`_CFRelease`, `_SCDynamicStoreCopyProxies`). Added `-framework CoreFoundation -framework SystemConfiguration` on the macOS vcpkg path. Both Linux arches remain green; this should complete the macOS link. 🙏 for another approval.

asubbarao:
Status update — all platforms now build cleanly except one transient infra hiccup:

- **Linux** (x64 + arm64): ✅
- **macOS** (arm64 + amd64): ✅  *(the earlier link gaps — `iconv`, then libcurl's `CoreFoundation`/`SystemConfiguration` proxy deps — are resolved)*
- **Windows** (x64-windows-static-release): ❌ **but not a code failure** — it's a transient vcpkg dependency download timeout:

  ```
  Downloading https://gitlab.freedesktop.org/.../freetype-VER-2-13-3.tar.gz
  error: ...freetype-VER-2-13-3.tar.gz: failed: status code 504
  ```

  `freetype` 504'd from the upstream mirror (cache miss → fallback download → gateway timeout). This commit (`03380d1`) only changed the **macOS** link branch — nothing Windows-related — and `freetype` built green on this same Windows runner in the immediately prior run.

Could you re-run the failed Windows job when you get a chance? (`gh run rerun 28227348408 --failed`, or the re-run button.) It should pick up the cached freetype on retry. Thanks!

carlopi:
Thanks! And welcome!

PR #2139, opened 2026-06-26
Update ducklink to v0.2.0

zacharywhitley:
## ducklink v0.2.0

Bumps the `ducklink` community extension from v0.1.0 to v0.2.0 (`repo.ref: v0.2.0`).

ducklink runs `duckdb:extension` WebAssembly **components** inside DuckDB — one
portable component, built once, runs on every platform without per-platform
native builds.

### What's new since v0.1.0
- **Versioned contract + load guard.** The `duckdb:extension` WIT contract is now
  versioned (`@2.0.0`) with a content-addressed identity (a `witcanon` digest of
  the contract shape). The host rejects components built against an incompatible
  contract at load time with an actionable message instead of silently
  mis-marshalling values.
- **Rich logical types.** Full marshalling for the expanded logical-type set —
  int8/16/32, the unsigned ints, float32, date/time/timestamp/timestamptz,
  interval, decimal, uuid, and the nested escape-hatch — across the component
  boundary.
- **wasmtime 46.0.1** (from the prior pin) and an embedded, vendored runtime so
  the build is self-contained.

`ducklink_version()` is built in, so the extension is usable and testable before
any component is configured:

```sql
LOAD ducklink;
SELECT ducklink_version();   -- 'ducklink 0.2.0'
```

Builds and tests green locally (`make release` + `make test_release` → SUCCESS).

### DIFF
--- a/extensions/ducklink/description.yml
+++ b/extensions/ducklink/description.yml
@@ -3,7 +3,7 @@
 extension:
   name: ducklink
   description: Run WebAssembly component extensions inside DuckDB
-  version: 0.1.0
+  version: 0.2.0
   language: Rust
   build: cargo
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

zacharywhitley:
I'm still working pretty heavily on it and I'm about to land some new features but I'm anxious to share it with duckdb and the community. I think it addresses a lot of the issues brought up in the presentation "DuckDB Extensions: The Past, the Present, and the Future". I'm currently writing a native pass through so you can load the webassembly version on any deployment, native, standalone wasm, or browser wasm but if you're on the native duckdb and install the native extension it will passthrough and you benefit from the native execution speed. With this you can rapid prototype with webassembly and get the security, stable api, etc and you get the capability rapidly. If there's demand you can transparently backfill a performant native implementation.

PR #2158, opened 2026-07-01
ducklink: update to v4.0.1 (SQL-native WebAssembly module loading)

zacharywhitley:
Updates `ducklink` from 0.5.2 to 4.0.0.

This release turns ducklink from a passive host (which registered only `ducklink_version()`) into a SQL-native loader for portable WebAssembly extension modules:

- **`ducklink_load('<name>')`** — resolve a module by name against a curated online catalog, download + sha256-verify + cache the component, and register its scalar/table/aggregate functions at runtime for use in later statements.
- **System views** — `ducklink.modules` (the catalog, with a `loaded` flag), `ducklink.functions` (every function with its full SQL signature), `ducklink.capabilities`, `ducklink.cache`, `ducklink.versions`.
- **`LOAD WASM '<name>'`** statement (Linux/macOS) — an explicit, sandbox-aware alternative to `ducklink_load`.
- `ducklink_version()` now carries a description.

The 4.0.0 version aligns ducklink with the `duckdb:extension@4` WebAssembly-component ABI generation it hosts (a lockstep-major policy: host major N runs module generation ≤ N).

Built on the Rust C Extension API. Windows builds ship the common tier (loader + views); the `LOAD WASM` *statement* is Linux/macOS-only because it binds DuckDB's internal C++ ABI, which has no Windows PE/COFF deferred-symbol equivalent — Windows users use `CALL ducklink_load(...)`, which is identical in effect.

### DIFF
--- a/extensions/ducklink/description.yml
+++ b/extensions/ducklink/description.yml
@@ -2,8 +2,8 @@
 # https://github.com/duckdb/community-extensions as extensions/ducklink/description.yml
 extension:
   name: ducklink
-  description: Run WebAssembly component extensions inside DuckDB
-  version: 0.5.2
+  description: Load portable WebAssembly extension modules into DuckDB from SQL
+  version: 4.0.0
   language: Rust
   build: cargo
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2161, opened 2026-07-01
zim: 0.4.1 → 0.5.0 (remote full-text search & suggest)

teaguesterling:
Bumps **zim** to v0.5.0.

Headline: **remote full-text search & suggest** — `zim_search`/`zim_suggest` now work over `http(s)`/`s3` URLs by reading the Xapian index in place via byte-range requests (a cold query on a 2.2 GB archive with a 166 MB index fetches ~0.3% of the file, not the whole index). Also: `with_snippet` toggle, honest `zim_info(...).has_fulltext_index`, and a `zim_remote_search_max_local_index` setting.

Implemented via a private vcpkg overlay of Xapian (adds `RandomAccessReader`) + libzim wiring; CI builds both from source on Linux/macOS/arm64/WASM with an http integration test.

- Release: https://github.com/teaguesterling/duckdb_zim/releases/tag/v0.5.0
- ref `87d627e` (main @ v0.5.0)

### DIFF
--- a/extensions/zim/description.yml
+++ b/extensions/zim/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: zim
   description: Read .zim (Kiwix / openZIM) archives directly in DuckDB via libzim, from local files or remote S3/HTTP — offline Wikipedia, WikiMed, Stack Exchange, iFixit, and more, with a zim:// filesystem and full-text search.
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2166, opened 2026-07-02
[ducksync] v0.2.0 -> v0.3.0 (Duckdb 1.5.4)

danjsiegel:
New version of DuckSync with three additions since the last community extension release:

- ducksync_serve and ducksync_stop for explicit Quack server lifecycle: users can now start a DuckDB Quack endpoint and connect DuckDB-native clients to it directly
- invalidation_mode = 'two_stage' on ducksync_create_cache: skips warehouse-backed freshness checks when SHOW TABLES says nothing changed, which cuts compute costs on stable tables
- Transparent cached-table routing via ReplacementScan: after a refresh, SELECT * FROM orders just works without needing to wrap everything in ducksync_query

Requires DuckDB v1.5.4+. No breaking changes to existing functions.


carlopi:
Thanks!

PR #2175, opened 2026-07-03
Fix bug on SQLite and enable WASM builds

SamArch27:
Hi there,

Would it be possible to also manually re-deploy the ADBC extension on v1.4.5 (Andium)?

Thanks!


carlopi:
Thanks!

PR #2178, opened 2026-07-05
Bump se3

jokasimr:
Bugfix:

* `null` in operations only invalidated top level struct, access to child entries returned non-`null`.

Documentation improvements.


carlopi:
Thanks!

PR #2182, opened 2026-07-06
lpts: bump to 1.1.0

ila:
Bumps the `lpts` extension to **1.1.0** (from 1.0.0).

- `version` → `1.1.0`, `repo.ref` → `6845654b3ce21922f83f226937abd3aa9a790a70` (the `v1.1.0` release commit on `cwida/lpts`).
- **Descriptor examples updated for the 1.1.0 API.** 1.1.0 removed `PRAGMA lpts_exec('query')` and the boolean-returning `PRAGMA lpts_check('query')`, replacing them with the `lpts_check` **session setting** (`SET lpts_check = true`, which transparently round-trip-checks every top-level `SELECT`). The `hello_world` and `extended_description` examples that used the removed pragmas would no longer run, so they now use `SET lpts_check = true` + a bare query.
- **`EXPLAIN (FORMAT SQL)` is now featured in `hello_world`** (previously only in the extended description).
- Dropped an incidental external reference.

### About 1.1.0

The release makes "LPTS never rewrites a query wrongly" an enforced invariant: a transparent round-trip checker plus a coverage gate that runs DuckDB's entire sqllogic corpus through LPTS (`files=3346`, `wrong=0 fail=0`). It also bumps the DuckDB target to `v1.5.4` and hardens non-DuckDB dialect output (Spark). Full notes: https://github.com/cwida/lpts/releases/tag/v1.1.0

All `hello_world` commands were verified against a local 1.1.0 build.

### DIFF
--- a/extensions/lpts/description.yml
+++ b/extensions/lpts/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: lpts
   description: Optimized-plan inspection and cross-system SQL transpilation
-  version: 1.0.0
+  version: 1.1.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2185, opened 2026-07-06
ai: bump to v0.4.0

leonardovida:
Bumps the `ai` extension to [v0.4.0](https://github.com/leonardovida/duckdb-ai/releases/tag/v0.4.0).

- Since 0.3.1: function descriptions/examples exposed via `duckdb_functions()` (so the generated docs tables on the extension page now include per-function descriptions and examples), a llama.cpp server provider, bind-verified SQL self-correction for the SQL assistant functions, and an embedding cache option fix


carlopi:
Thanks!

PR #2201, opened 2026-07-08
ducklink: update to v4.1.1

zacharywhitley:
Additive follow-up to v4.1.0 (currently open as [#2199](https://github.com/duckdb/community-extensions/pull/2199)). v4.1.1 lands the **wasm custom-section docs bridge**: components can ship their own `summary` / `description` / `example` / `tags` inside the `.wasm` binary via a `duckdb.docs` custom section, and ducklink merges them with the catalog docs at query time.

### Merge semantics

Per-function, per-field:
- Component `summary` / `description` / `example` **replace** the catalog value when present.
- `tags` **union** with the catalog set (catalog order preserved, component-only tags appended).
- Modules without a section render pure catalog data — no behaviour change for existing components.

### Why it exists

Catalog authoring lag is a real thing. This lets third-party component authors keep their own docs authoritative without waiting on a catalog PR to land, while still respecting the catalog as source-of-truth for anything the component doesn't override.

### Safety

Non-fatal at every step: missing file / malformed envelope / missing section / invalid JSON / wrong shape all return silently. A broken section can never break `LOAD WASM 'name'`. Diagnostics go through `verbose_log!`; no user-facing noise on the happy path.

### Implementation

- New module `src/docs_section.rs` (~260 lines, unit-tested).
- Hand-rolled uleb128 wasm scanner — no new dependencies. Walks top-level custom sections plus one level into component-nested core modules.
- Wired through `Engine2::load` + `build_doc_rows`, so the override flows through `ducklink.docs`, `ducklink_search`, and `ducklink_help` uniformly.
- 32 lib tests pass; release build clean.

### PR history

Supersedes:
- [#2199](https://github.com/duckdb/community-extensions/pull/2199) (v4.1.0, still open) — same code plus the wasm-section reader.
- [#2195](https://github.com/duckdb/community-extensions/pull/2195) (v4.0.2, closed).
- [#2192](https://github.com/duckdb/community-extensions/pull/2192) (v4.0.1, closed; transient CI failure on osx_arm64).

Closing #2199 alongside this PR with a supersedes comment.

PR #2207, opened 2026-07-09
ducklink: update to v4.2.0

zacharywhitley:
Six perf batches + a cold-start UX fix + an optional binary-size feature since v4.1.1. No API breaks, no WIT contract changes — additive minor bump.

## User-visible wins

Measured on darwin/aarch64 with lto=true, strip=true:

| Metric | Before | After | Δ |
|---|---:|---:|---:|
| Aggregate 1M rows (`sample_sum(v) FROM range(1M)`) | 82 ms | 38 ms | **-54%** |
| Parallel scalar (2-thread cross-extension) | 12.9 ms | 5.75 ms | **-55%** |
| Serial scalar dispatch | 42 ms | 11.2 ms | **-73%** |
| Cold-start on offline / bad-DNS | 15 s | 3 s | **-80%** |
| Binary size (offline build variant) | 25.9 MB | 21.7 MB | **-4 MB** |

## What's new architecturally

**Dispatch (F/G/H/I/L batches)**
- Column-native aggregate UPDATE: `AggState` is per-column typed accumulators picked at `agg_init`. Ungrouped chunks bulk-copy the whole input via `extend_from_slice`. Eliminates ~10M outer Vec allocs + boxed enum variants per 10M-row aggregate.
- `CallbackRegistry` `Mutex → RwLock`; `Weak<Instance>` on `CallbackEntry` so dispatch prologue is one atomic upgrade instead of `HashMap<name, _>::get` + `Arc` clone.
- Per-thread `SCALAR_ARGS_SCRATCH`: `refill_colvec` reuses Colvec buffers across chunks; TEXT/BLOB reuse via `String::clear + push_str`.
- `write_colvec` skips a redundant 2048-byte null-mask scan.

**Discovery / docs (F1, F2, G4)**
- `Arc<CachedDocs>` with generation-counter invalidation for the docs pipeline. `render_help`, `ducklink.docs`, and `ducklink_search` no longer rebuild ~639 DocRows per query.
- `flat_vector` handles hoisted out of every discovery TF's row loop.

**Load-time (H4, J3)**
- Base component-linker template cached per Engine: ~25 `add_to_linker` calls run ONCE per process instead of once per component load.
- Wasm bytes read ONCE per load and shared between `Component::from_binary` and `parse_docs_from_bytes`.
- `wasmtime::Config::memory_may_move(false)` (K1): 1-2% real per-row gain via base-pointer LICM.

## Cold-start UX (J1)

The single most user-visible slow path in the codebase: opening DuckDB on hotel-WiFi / offline, running `SELECT * FROM ducklink.modules`, and staring at a black cursor for 15 seconds while a network fetch times out before the bundled snapshot loads.

- Catalog HTTP fetch timeout: **15 s → 3 s**. The graceful fallback (bundled snapshot) is already in the binary.
- Async prewarm: `Engine2::new` spawns a background thread that populates the catalog `OnceLock` while DuckDB is still finishing extension setup. By the time the first user query arrives the catalog is usually already resolved. Best-effort; races are resolved via `OnceLock::get_or_init`'s barrier.

## Optional binary-size feature (J2)

New default-enabled `network` feature gates `reqwest + rustls + webpki-roots + sha2`:

```
cargo build --release                                     # 25.9 MB (with network)
cargo build --release --no-default-features \
            --features loadable,advanced                  # 21.7 MB (offline)
```

The offline variant is for air-gapped / minimum-size deployments; it serves only names in the bundled snapshot and refuses to download uncached blobs.

**Community-extensions CI runs `cargo build --release` (default features)** so this PR's distributed artifact keeps full network support — no change for end users of the CI-built extension.

## Retracted design docs

`docs/perf-ceiling-measurement.md`, `docs/wit-streaming-scalar-dispatch.md` (RETRACTED), `docs/wit-shared-memory-result.md` (CANCELLED), and `docs/guest-simd-investigation.md` capture the measurement work behind this release and the reasoning behind rejecting several proposed architectural changes (~30% streaming projection turned out to be ~1.3%; ~27% zero-copy projection turned out to be <1%). New benches (`streaming_validation`, `memcpy_floor`) let anyone reproduce the findings.

## PR history

Supersedes v4.1.1 (merged as #2201). All changes are on top of `v4.1.1`.

### DIFF
--- a/extensions/ducklink/description.yml
+++ b/extensions/ducklink/description.yml
@@ -3,7 +3,7 @@
 extension:
   name: ducklink
   description: Load portable WebAssembly extension modules into DuckDB from SQL
-  version: 4.0.0
+  version: 4.2.0
   language: Rust
   build: cargo
   license: MIT
@@ -19,7 +19,7 @@ extension:
 
 repo:
   github: tegmentum/ducklink-extension
-  ref: v4.0.0
+  ref: v4.2.0
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


zacharywhitley:
Superseded by #2214 (v4.3.0).

PR #2210, opened 2026-07-09
Add splunk extension

smithclay:
This adds an extension that reads logs from a Splunk instance's search REST API and hands them back as ordinary DuckDB rows.

The rows follow the [OTLP logs schema](https://smithclay.github.io/duckdb-otlp/reference/schemas/#logs-read_otlp_logs), the same columns that `read_*_logs` produces in the `otlp` and `datadog` extensions. Someone keeping logs from several systems in one lakehouse can query them together and stop thinking about where each one came from.

Example query:

```
SELECT time_unix_nano, service_name, severity_text, body 
FROM read_splunk_logs(query => 'index=main error', earliest => '-1h');
```

### DIFF
--- a/extensions/splunk/description.yml
+++ b/extensions/splunk/description.yml
@@ -1,0 +1,65 @@
+extension:
+  name: splunk
+  description: Read logs from a Splunk search REST API into OTLP-shaped tables
+  version: 0.1.0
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2199, opened 2026-07-08
ducklink: update to v4.1.0

zacharywhitley:
Updates `ducklink` from v4.0.0 (upstream) to v4.1.0. Supersedes [#2195](https://github.com/duckdb/community-extensions/pull/2195) (v4.0.2), which never merged; v4.1.0 rolls the intervening in-database documentation surface + catalog authoring guide into a single fresh submission alongside all of the v4.0.2 perf and discovery-view work.

## In-database documentation (new in v4.1.0)

Users of the extension shouldn't need to leave DuckDB to know what a WebAssembly module in the catalog does. v4.1.0 lands the doc / search / help surface — three coordinated tools plus the catalog schema they share:

- `ducklink.docs` (view over `ducklink_docs()` TF) — flat one-row-per-function surface, columned as (module, function, kind, signature, summary, description, example, tags, loaded). For plain WHERE-clause discovery: `SELECT * FROM ducklink.docs WHERE description ILIKE '%routing%';`
- `ducklink_search('query')` — ranked matches. Tokenizes on whitespace (case-insensitive substring per token) and scores 10x name + 5x tags + 3x summary + 1x description, filters `score > 0`, orders stable by (score DESC, module, function).
- `ducklink_help('name')` — pretty-printed markdown for a single function OR every function in a module; a not-found hint points at `ducklink_search`.

The underlying `CatalogEntry.functions[]` gains four optional serde-default fields — summary, description, example, tags — so unenriched catalog JSON parses unchanged. The bundled snapshot's `aba` and `sample_extension` entries ship enriched so the pipeline demos end-to-end.

A companion `docs/catalog-authoring.md` guide walks catalog authors through enriching `functions[]` for the doc surface, including the 10x/5x/3x/1x ranking weights and how to preview edits locally.

## Perf pass (from v4.0.2, still not merged upstream)

Measured medians on aarch64-darwin, v4.0.1 baseline -> v4.0.2:

- `aggregate_query/sample_sum_1M`: 76.2 ms -> 73.3 ms (**-3.7%**)
- `scalar_dispatch/plus_one_i64_2048`: 139.8 us -> 134.3 us (**-5.3%**)
- new `scalar_dispatch/plus_one_col_i64_2048`: 86.4 us — the col-native path is **~35% faster** than the row-major sibling
- `parallel_scalar/parallel_cross_ext`: flat within CI after cache-line-padding the callback registry

[...]

PR #2195, opened 2026-07-08
ducklink: update to v4.0.2

zacharywhitley:
Updates `ducklink` from v4.0.0 (upstream) to v4.0.2. Supersedes [#2192](https://github.com/duckdb/community-extensions/pull/2192) (v4.0.1); that PR merged but the post-merge build hit a transient `vcpkg` download failure on `osx_arm64` and the CDN kept serving v0.5.2. Rather than a re-run, v4.0.2 rolls the intervening perf pass + discovery-view UX cleanup into a fresh submission.

## Perf pass

Measured medians on aarch64-darwin, v4.0.1 baseline → v4.0.2:

- `aggregate_query/sample_sum_1M`: 76.2 ms → 73.3 ms (**−3.7%**)
- `scalar_dispatch/plus_one_i64_2048`: 139.8 µs → 134.3 µs (**−5.3%**)
- new `scalar_dispatch/plus_one_col_i64_2048`: 86.4 µs — the col-native path is **~35% faster** than the row-major sibling
- `parallel_scalar/parallel_cross_ext`: flat within CI after cache-line-padding the callback registry

Highlights: column-native scalar and aggregate dispatch (skip the runtime's row-major pivot on both sides of the WIT crossing), per-instance locking so cross-extension parallelism doesn't serialize on one mutex, and the advanced-tier pushdown scan now writes column-hoisted.

## Discovery views

Renamed for readability (SQL surface change, no wasm ABI impact — a v4.0.x host still loads every v4.0.x guest):

- `ducklink.versions` → `ducklink.module_compatibility`
- `ducklink.capabilities` → `ducklink.host_capabilities`
- new `ducklink.host` view — this host's WIT contract version + DuckDB build info
- `ducklink.modules.language` dropped (source language isn't useful — everything is wasm at runtime)
- `ducklink.modules.requires` → `capabilities` (pairs with `host_capabilities`: host provides, module uses)

Manifest `extended_description` updated to match.

PR #2192, opened 2026-07-07
ducklink: update to v4.0.1

zacharywhitley:
Re-triggers the community build. The prior v4.0.0 post-merge build ([#2158](https://github.com/duckdb/community-extensions/pull/2158), [run 28497377246](https://github.com/duckdb/community-extensions/actions/runs/28497377246)) failed on `osx_arm64` with a transient vcpkg download failure (`curl (6) Could not resolve host: github.com`), so the CDN is still serving v0.5.2 and `SELECT * FROM ducklink.modules` fails against the community-installed extension.

v4.0.1 also folds in three fixes on top of v4.0.0:
- `feat(events)`: add the `ducklink.events` runtime audit-log system view
- `feat(compat)`: enforce strict same-major loading and re-embed the gen-4 snapshot
- `chore`: point community manifest ref at v4.0.1

PR #2214, opened 2026-07-09
ducklink: update to v4.3.0

zacharywhitley:
Additive minor since v4.2.0. No API breaks, no WIT contract changes.

## What's new

**Native distribution end-to-end.** A new `native` provider kind on ducklink catalog entries (platform + duckdb_version + content_digest). Entries with a matching native provider are fetched from ducklink infrastructure, sha256-verified, cached under `~/.cache/ducklink/native/…`, and LOADed via DuckDB's own extension mechanism. The extension does **not** flip `allow_unsigned_extensions` — that's an explicit user choice at DuckDB startup (`-unsigned`).

**New SQL surface.** `DUCKLINK LOAD '<name>' [WASM|NATIVE];` — default WASM. Retires the previous `LOAD WASM 'aba'` / `LOAD NATIVE 'aba'` pseudo-LOAD verbs (which borrowed DuckDB's LOAD keyword; the new form is unambiguously ducklink).

**Function form.** `FROM ducklink_load('aba', kind => 'wasm' | 'native')` — same choice via the existing table function.

**Discovery.** New `native_available` BOOLEAN column on `ducklink.modules` — TRUE when the entry has a native provider matching this host's platform + DuckDB version.

## Design docs shipped alongside

Two upstream discussion drafts are in the repo, both ready to file at duckdb/duckdb:

- `docs/duckdb-upstream-custom-trusted-keys.md` — minimal-viable subset of [#23388](https://github.com/duckdb/duckdb/discussions/23388); a `custom_trusted_extension_keys` setting so third-party extension distributors can be trusted without global `allow_unsigned_extensions`.
- `docs/duckdb-upstream-function-autoload.md` — an `ExtensionHelper::RegisterFunctionEntry` API so extensions like ducklink can extend DuckDB's function → extension autoload map at runtime. Would enable transparent autoload (`SELECT aba_validate(...)` with no prior LOAD).

## Supersedes

[#2207](https://github.com/duckdb/community-extensions/pull/2207) (v4.2.0). Closing that with a supersedes comment.

## PR history

- #2207 v4.2.0
- #2201 v4.1.1
- #2199 v4.1.0
- #2195 v4.0.2
- #2192 v4.0.1

### DIFF
--- a/extensions/ducklink/description.yml
+++ b/extensions/ducklink/description.yml
@@ -3,7 +3,7 @@
 extension:
   name: ducklink
   description: Load portable WebAssembly extension modules into DuckDB from SQL
-  version: 4.0.0
+  version: 4.3.0
   language: Rust
   build: cargo
   license: MIT
@@ -19,7 +19,7 @@ extension:
 
 repo:
   github: tegmentum/ducklink-extension
-  ref: v4.0.0
+  ref: v4.3.0
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2218, opened 2026-07-10
pdf: bump to 0.5.0

asubbarao:
Bumps the `pdf` extension 0.3.0 → 0.5.0, pinned to [`9006e93`](https://github.com/asubbarao/duckdb-pdf/commit/9006e9388fa5d7c8e5e1816ac4e9a7a6e5b6e8e0) (all-platform CI green on the extension repo, including windows_amd64/MSVC).

New since 0.3.0 — the extension grows from a reader into a full PDF toolkit:

- **Structure & RAG**: `read_pdf_elements` (layout elements with bounding boxes, deterministic geometry), `pdf_chunks` (heading-aware retrieval-ready chunking).
- **Inspection**: `pdf_info`, `pdf_outline`, `pdf_attachments`, `pdf_form_fields`, `pdf_annotations`, `pdf_revisions` (incremental-update forensics), `pdf_signatures` (detection + cryptographic integrity verification over signed byte ranges), `pdf_images` (embedded raster extraction).
- **Transform (qpdf)**: `pdf_merge` / `pdf_split` / `pdf_split_blank` / `pdf_rotate` / `pdf_pages` / `pdf_compress` / `pdf_encrypt` / `pdf_decrypt` (AES-256 R6) / `pdf_watermark` / `pdf_bates`.
- **Robustness**: `ignore_errors` on `read_pdf` / `read_pdf_meta` for fault-isolated folder scans.

qpdf (12.3.2, vcpkg, default-features=false) is compiled in an isolated C++17 translation unit so the extension still builds against the registry's default C++11 driver. Descriptor docs updated to cover the new surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2224, opened 2026-07-11
pdf: 0.5.1 — fix ref, 0.5.0 was pinned to a commit missing all its features

asubbarao:
0.5.0 (#2218) was pinned to `ref: 9006e93`, which is the batch-1 baseline — it **predates every feature the 0.5.0 description advertises**. At that commit `pdf_images`, `pdf_split_blank`, `pdf_watermark`, `pdf_bates`, `pdf_redact`, and `pdf_sign` don't exist, so a user who installs 0.5.0 and calls any of them gets `Catalog Error: function does not exist`. My apologies for the bad ref in #2218 — the description was written for the feature surface but the ref pointed at the last commit that had cleared all-platform CI at the time.

This bumps the ref to [`177b634`](https://github.com/asubbarao/duckdb-pdf/commit/177b634edbcf7e3fdbd7068f4a72455073fba67c) (duckdb-pdf `main`), where all of those functions are present and tested, plus:

- **auto-OCR is now best-effort** — a blank/scanned page encountered during a plain `read_pdf` no longer hard-errors on machines without a Tesseract model installed; it degrades to empty text for that page. Explicit `ocr := true` still raises the actionable install error. (Without this, the extension's own test suite failed on the registry's model-less Linux runner.)
- **`pdf_revisions` no longer fatally invalidates the connection** on a header-less input — an unescaped `%PDF-` in the error format string had been escalating a clean `IOException` into a fatal `InternalException`.
- **Windows-portable test assertions** — two feature tests asserted a forward-slash path (Windows glob expansion returns backslashes) and trimmed `' \\n\\f'` but not the `'\\r'` in Windows line endings; both fixed (test-only). This is what an earlier ref (d33b1f3) still tripped on the windows_amd64/MSVC leg.

Only the pinned ref changed vs the merged 0.5.0; the description already documents these functions.

### DIFF
--- a/extensions/pdf/description.yml
+++ b/extensions/pdf/description.yml
@@ -17,7 +17,7 @@ extension:
 
 repo:
   github: asubbarao/duckdb-pdf
-  ref: d33b1f327ddca3ba963746dc2ecce6e1c804f01b
+  ref: 177b634edbcf7e3fdbd7068f4a72455073fba67c
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks!

PR #2213, opened 2026-07-09
Add gorz extension

gorfather:
Adds **gorz** — read GORpipe `.gorz` and `.gord` files and write `.gorz` as native DuckDB tables.

GORpipe (https://github.com/gorpipe/gor) is a genomic analysis engine; `.gorz` is its block-compressed, genomically-ordered table format and `.gord` a dictionary of partition files. This extension exposes them to DuckDB:

- `read_gor` / `read_gorz` / `read_gord` table functions (projection + parallel scan, `WHERE chrom/pos` → block seek, `range := 'chrN:a-b'` GOR `-p` filter, `f`/`ff` partition-tag filters).
- Replacement scan, so `FROM 'x.gorz'` works directly.
- `COPY (…) TO 'x.gorz' (FORMAT gorz)` — write path (validates GOR order).
- Object stores via DuckDB's FileSystem (`s3://…` with httpfs).

Files written are read/seeked natively by `gorpipe` itself, and vice-versa.

- Repo: https://github.com/gorfather/duckdb-gorz @ v0.1.0
- Language: C++ · Build: cmake · License: MIT
- Built + tested against DuckDB v1.5.x (`make test`: 29/29 assertions pass).

PR #2225, opened 2026-07-11
gorz: bump to 0.1.1 (ff as tag file, .gord source-name parity)

gorfather:
Follow-up to the merged v0.1.0 (#2213). Bumps `gorz` to **0.1.1** and points the ref at [gorfather/duckdb-gorz@582880d](https://github.com/gorfather/duckdb-gorz/commit/582880d0f5ab23423f6680fcd82babb4fce1c4e1).

### Changes
- **`ff` is now a tag file**, distinct from `f`. `f := ['A','B']` is an inline list; `ff := 'tags.txt'` is a path to a tag file (one tag per line / first tab-delimited column, `#` lines skipped), read via DuckDB's FileSystem (object stores work). Previously both were lists.
- **`.gord` source-column name parity with GORpipe** — the dictionary header is split on tab or comma, the source name is its 2nd column, and a dummy `col2` header is ignored (falls back to `Source`).

### Testing
Built against DuckDB v1.5.4 through the standard vcpkg+ninja toolchain; `test/sql/gord.test` (now covering `ff`-file + `f`/`ff` union) and `gorz.test` pass — 88 assertions. Locally verified on osx_arm64.

### DIFF
--- a/extensions/gorz/description.yml
+++ b/extensions/gorz/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: gorz
   description: Read GORpipe .gorz and .gord files and write .gorz as native DuckDB tables
-  version: 0.1.0
+  version: 0.1.1
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


gorfather:
Hi @samansmink — CI is green across the full matrix here. This is the v0.1.1 follow-up to #2213 (ff becomes a tag file rather than a duplicate of f, plus .gord source-column-name parity with GORpipe). Ready to merge whenever you have a moment — thanks!

PR #1915, opened 2026-05-14
Update duck lineage to 0.2.1

thijs-s:
This release brings improvements to robustness and some additional noise-filtering.

PR #2241, opened 2026-07-14
Fix YAML frontmatter generation for multi-line descriptions

Flogex:
The page of the duck_lineage community extension is broken because the front matter is not indented correctly, see
https://duckdb.org/community_extensions/extensions/duck_lineage and
https://github.com/duckdb/duckdb-web/blob/3eb8acd5be7c58ff7ef8a190059a7292be9cc21a/community_extensions/extensions/duck_lineage.md. The reason for this is that #1915 added a literal block with multiple lines as description.

I modified _scripts/generate_md.sh_ to add correct indentation (two spaces) for every line, not just the first one.

The new output looks like this:
```yaml
excerpt: |
  DuckDB Community Extensions
  DuckLineage automatically captures data lineage from every DuckDB query and emits OpenLineage
  events to any OpenLineage-compatible backend such as Marquez, DataHub,or Atlan.
  Includes column-level lineage, schema facets, DuckLake catalog support, asynchronous
  HTTP delivery, and parent-run linking for orchestrators like Airflow and Dagster.
  Built and maintained by Ilum Labs.
```

An alternative would be to enforce using folded block scalars (`>`) in YAML, but letting the _generate_md.sh_ script handle this is the more general approach.

Also fixes the GitHub URL.


sebastiaan-dev:
Thanks!

Issue #2234, opened 2026-07-13
Clarify support for versioned community extension releases


d33bs:
Hi,

I’m trying to understand the intended path for making multiple versions of the same community extension available to users.

For example, if an extension has upstream releases `v0.1.0` and `v0.1.1`, the community descriptor currently has a single `repo.ref`, so updating it appears to replace the active version used for future builds.

Is there a supported way for users to install a specific community extension version, or should maintainers publish older versions through a separate custom extension repository/artifact path?

A few related questions:

- Is `deploy_versioned` intended to support this use case?
- If versioned community extension installs are not supported, is the recommended approach to use separate repositories/channels?
- Would it make sense to document the expected workflow for maintainers?

Thanks in advance for your help!

rustyconover:
Installing a specific extension version isn't possible in DuckDB at this time because each release overwrites the extension path.

d33bs:
Thanks @rustyconover , that makes sense.

So for community extensions, updating `repo.ref` should be treated as replacing the active artifact for a given DuckDB version/platform, and maintainers should not expect users to be able to select older extension versions through `INSTALL ... FROM community`.

Would a small docs note be welcome to make that explicit for maintainers? Something like:

> Community extension descriptors point to a single active source ref. Rebuilding an extension overwrites the artifact path for each DuckDB version/platform, so DuckDB does not currently support installing a specific historical community extension version. Maintainers who need version-pinned access should publish separate custom repositories or direct artifacts.

Separately, would maintainers be open to exploring version-addressable extension artifacts in the future, or is that outside the intended scope for community extensions? It'd be amazing to avoid having to manage this independently.

rustyconover:
A docs note along those lines would definitely help — that matches how `INSTALL ... FROM community` resolves today: one active ref per DuckDB version/platform, and a rebuild replaces the artifact in 
place, so there's no way to address an older build.  But DuckDB could do better if they chose to do so.

On the version-addressable question: this is exactly the itch we ([Query Farm](https://query.farm)) scratched in [Haybarn](https://github.com/Query-farm-haybarn) (our Query Farm's downstream DuckDB-compatible distribution), so I'll share the shape in case it's a useful data point for what a future design could look like. The core change was making the extension storage layout append-only instead of a single mutable slot like DuckDB does.

The R2 (Cloudflare's S3-like service) layout is keyed by the extension's **source commit**:

```
# immutable — one directory per build, never overwritten
[...]

PR #2243, opened 2026-07-14
Clarify version-pinning behavior for community extensions

d33bs:
This PR updates UPDATING.md to clarify that community extension descriptors point to a single active repo.ref. When an extension is rebuilt, the artifact for each DuckDB version/platform is replaced at the standard install path, so DuckDB does not currently support installing a specific historical community extension version through the normal community extension flow.

It also notes that maintainers who need version-pinned or historical access should publish builds through separate custom repositories, separate channels, or direct artifact URLs.

Executes on some of the conversation found in #2234


samansmink:
looks good, thanks @d33bs!

d33bs:
Thanks @samansmink !

PR #2268, opened 2026-07-19
Add trino_parity extension

apatrida:
This PR adds `trino_parity`, a C++ extension that provides native
`trino_<name>(...)` scalar functions for the specific cases where DuckDB's
built-ins diverge from Trino's documented behaviour on Unicode and byte-level
inputs.

It is intended to be loaded server-side by anything that pushes Trino-shaped
predicates down to DuckDB, so those predicates evaluate with Trino semantics
instead of silently dropping rows on non-ASCII input.

### What it ships (10 native functions + `trino_meta()`)

- **String (statically-linked ICU):** `trino_lower`, `trino_upper`,
  `trino_reverse`, `trino_trim`, `trino_ltrim`, `trino_rtrim`,
  `trino_normalize` — root-locale full case folding (`'İ'`→`'i'`+U+0307,
  `'ß'`→`'SS'`), code-point reverse, `Character.isWhitespace` trim, NFC.
- **Hash (vendored single-file libs):** `trino_xxhash64`, `trino_sha512`,
  `trino_hmac_sha256` over raw `VARBINARY` — no runtime dependency on other
  extensions.

Aligned Trino functions (`length`, `abs`, `year`, `substring`, …) are
intentionally not shipped; a caller emits those as bare DuckDB SQL.

### Build / CI

- Standard extension-ci-tools toolchain, `build: cmake`, license MIT.
- Builds green on the full matrix against DuckDB v1.5.4: Linux amd64/arm64,
  macOS amd64/arm64, Windows (MSVC + MinGW), and DuckDB-Wasm (mvp/eh/threads),
  plus Format + Tidy checks.
- ICU is vendored under `third_party/icu/`; no extra vcpkg dependencies.

Repo: https://github.com/brikk/duckdb-trino-parity-extension (ref pinned in
`description.yml`, v0.2.0).

### DIFF
--- a/extensions/trino_parity/description.yml
+++ b/extensions/trino_parity/description.yml
@@ -1,0 +1,77 @@
+extension:
+  name: trino_parity
+  description: Native Trino-compatible scalar functions (trino_*) for the cases where DuckDB's built-ins diverge from Trino on Unicode and byte-level inputs — ICU full case folding, code-point reverse, Java-whitespace trim, NFC normalize, and raw-bytes xxhash64/sha512/hmac_sha256
+  version: 0.2.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - apatrida
+
+repo:
+  github: brikk/duckdb-trino-parity-extension
+  ref: 8527a7fc00e2a3baf117cbf641b973d7cd9d931e
+
+docs:
+  hello_world: |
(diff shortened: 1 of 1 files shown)


apatrida:
This is used primary by trino-duckbridge (quack support, pushdown support, ...)
https://github.com/brikk/duckbridge

PR #2284, opened 2026-07-22
Bump plinking_duck to v0.7.2

teaguesterling:
Updates plinking_duck to v0.7.2 (ref 0a37fcc).

Highlights since v0.7.1:
- read_pfile honors `file_search_path` and expands globs / `pathmacro:` URLs to local shard paths
- multi-file `orient := 'sample'` + new `combine_samples` parameter (implicit/identical)
- shared `psam` override allowed with a multi-file list
- streaming per-sample `genotypes := 'counts'|'stats'` (O(samples) memory, not the variants×samples matrix) + optional sparse (pgen difflist) path for rare-variant carrier finding (`plinking_sample_counts_sparse`, ~4-6x)
- fix: crash on a region matching zero variants via parquet pushdown

Full suite green (74 cases / 4189 assertions).

### DIFF
--- a/extensions/plinking_duck/description.yml
+++ b/extensions/plinking_duck/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: plinking_duck
   description: Read PLINK 2 genomics file formats and run common genetic analyses directly in SQL
-  version: 0.7.1
+  version: 0.7.2
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


teaguesterling:
Superseded by #2298 (v0.8.0), which includes all of v0.7.2 plus remote/VFS .pgen reads. Closing this in favor of the v0.8.0 bump.

PR #2348, opened 2026-07-25
duck_block_utils: 1.3.0 (terminal rendering + DuckDB v1.5.5)

teaguesterling:
Bumps `duck_block_utils` to **1.3.0**, `ref: refs/tags/v1.3.0`.

## What's new
- **ANSI terminal rendering** — `PRAGMA duck_block_render`, width-aware C++ `db_blocks_render_ansi`, and `db_render_blocks` / `db_render_block` / `db_render_query`.
- **Page composition macros** — `db_table`, `db_query_table` (embed query results), `db_page`.
- **Pandoc-AST** recursion-depth cap + fidelity improvements.
- **Retargeted** the stable build to the **DuckDB v1.5.5** release tag (v1.5-variegata line).

Upstream CI is green across Linux amd64/arm64, macOS amd64/arm64, and all Wasm legs; full test suite passes (538 assertions).

## andium (v1.4.5 track)
Intentionally left at the v1.2.1 commit: v1.3.0 is a DuckDB v1.5.5 tree that dropped v1.4.x support and is not built-verified against v1.4.5 — same pattern as `func_apply`. v1.4.5 users stay on the existing track; v1.5.x users get 1.3.0 via `ref`.

### DIFF
--- a/extensions/duck_block_utils/description.yml
+++ b/extensions/duck_block_utils/description.yml
@@ -143,6 +143,32 @@ docs:
     | `pandoc_inlines_to_db_inlines(json)` | Parse Pandoc inline AST |
     | `db_inlines_to_pandoc(blocks)` | Convert to Pandoc inline AST |
 
+    ## Terminal Rendering
+
+    Render documents — or any query result — as styled ANSI terminal output
+    (`PRAGMA duck_block_render`), and compose pages that embed live query
+    results:
+
+    ```sql
+    PRAGMA duck_block_render;
+
+    -- Any query as a pretty ANSI table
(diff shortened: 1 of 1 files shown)


teaguesterling:
Holding this release. v1.3.0's ANSI renderer re-parses Markdown syntax out of block `content` and skips structured `kind='inline'` elements — that bakes one downstream format's syntax into the format-agnostic core and violates the duck_blocks spec (rich text is structured inline children; `content` is literal per `encoding`). Will resubmit once the renderer is format-agnostic. Sorry for the churn.

PR #2363, opened 2026-07-26
Add motorsport_telemetry extension

tobi:
Adds `motorsport_telemetry`, a Rust extension that queries motorsport data-logger files directly in DuckDB:

- Pi Research / Cosworth **PDS** (`.pds`)
- MoTeC i2 **LD** (`.ld`)
- Racelogic VBOX **VBO** (`.vbo`)

The model is two relations per file — channel metadata and native-rate samples — plus an interpolated wide reader. Files are memory-mapped, values decode straight into DuckDB vectors, scans are parallel, and projection pushdown skips channels a query does not touch. Integer and event channels always use previous-value semantics; only floating-point channels interpolate linearly.

```sql
SELECT name, unit, frequency_hz, sample_count
FROM telemetry_metadata('run.ld')
WHERE sample_count > 0;

SELECT *
FROM read_telemetry('run.ld',
  rate := 100,
  channels := 'Ground Speed,Throttle Pos,Brake Pedal Pos');
```

Notes for reviewers:

- Uses only the stable public C extension API; built with `extension-ci-tools` (`make configure/release/test_release`).
- Repository CI builds through the Community Extension toolchain against DuckDB 1.5.4 and runs parser, integration, and registration tests on Linux, macOS, and Windows.
- All test fixtures are generated at runtime, so no telemetry data is shipped.
- `excluded_platforms` skips DuckDB-Wasm, musl, MinGW, and RTools for the initial submission; a separate WASM build is published from the project's own repository.
- Pinned ref `b0a16ce6f79af7672a031524132414b848de8ff8` is tag `v0.6.0`.

Source: https://github.com/tobi/duckdb_motorsport_telemetry

### DIFF
--- a/extensions/motorsport_telemetry/description.yml
+++ b/extensions/motorsport_telemetry/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: motorsport_telemetry
   description: Query Cosworth PDS, MoTeC LD, and Racelogic VBOX telemetry directly in DuckDB
-  version: 0.6.0
+  version: 0.6.1
   language: Rust
   build: cargo
   license: MIT
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: tobi/duckdb_motorsport_telemetry
-  ref: b0a16ce6f79af7672a031524132414b848de8ff8
+  ref: 8cc06a597e33e7558be759cbffa44f2a2e1baa49
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


tobi:
The `linux_amd64` failure was on our side. Our `test_release` target depended on `release`, so when the workflow ran `make test_release` on the host after building inside the container, cargo tried to rebuild against a `target/` directory owned by root and failed with `error: failed to open: target/release/.cargo-build-lock (Permission denied)`. `test_release` now only runs the tests, matching the standard layout.

Descriptor updated to version `0.6.1`, ref `8cc06a597e33e7558be759cbffa44f2a2e1baa49` (tag `v0.6.1`).

PR #2394, opened 2026-07-28
laterite_ags4: bump to 0.9.0

niko86:
Bumps `laterite_ags4` to **0.9.0**.

This PR was opened for 0.8.0 and has not been merged, so rather than open a
second one it has been re-pointed at 0.9.0 — the branch name still reads
`laterite_ags4-0.8.0`, but the descriptor and the pinned ref are 0.9.0. That
also means this takes the registry **0.7.0 → 0.9.0** directly; 0.8.0 was never
published here.

`ref` is `01ea351ba30e2ec1dfb75f834be2a07d287d40e8` — a tagged release
(`v0.9.0`) of niko86/laterite-duckdb, built and tested in that repo's CI.

## What changed since 0.7.0

Nothing in the extension's SQL surface is removed or renamed. The two breaking
changes in the 0.9.0 engine release are on other surfaces (a browser/wasm API,
and file *emission*); this extension is a read-only reader and takes the engine
core with `default-features = false`, so neither reaches it.

The `docs/function_descriptions.csv` row for `ags_rules()` is included — it
shipped earlier but had never reached the published docs page, which renders
from both files.

### DIFF
--- a/extensions/laterite_ags4/description.yml
+++ b/extensions/laterite_ags4/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: laterite_ags4
   description: Read AGS4 geotechnical data files as typed, UUID-keyed tables directly from SQL — born-typed columns, deterministic content-addressed keys that join across groups by construction, and an embedded AGS dictionary. A read-only SQL surface. Local, http(s):// and s3:// (with httpfs).
-  version: 0.7.0
+  version: 0.9.0
   language: Rust
   build: cargo
   license: MIT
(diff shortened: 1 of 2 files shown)


niko86:
Closing as a duplicate of #2411 — please review that one instead.

Both PRs carry byte-identical changes (`laterite_ags4` 0.9.0, pinned to `01ea351ba30e2ec1dfb75f834be2a07d287d40e8`). This PR was originally opened for 0.8.0 and I re-pointed it at 0.9.0 by hand, not realising our release workflow already opens a fresh per-version PR on tag — which is #2411.

#2411 is the one to keep: its branch (`laterite_ags4-0.9.0`) matches its contents, and our automation keeps that branch in sync on any re-tag, whereas this one would go stale.

Apologies for the duplicate review surface.

PR #2399, opened 2026-07-29
Add quackiso

tempoloss:
Adds `quackiso` — a Rust extension that queries ISO 20022 financial messages as SQL.

v1 ships `read_iso20022(path)` for camt.053 bank statements, flattening each booked entry into one row (amount, currency, credit/debit, status, dates, counterparty, end-to-end id, remittance). The reader streams one `<Ntry>` subtree at a time, so a multi-GB statement parses in constant memory (verified: 1.7 GB / 3M entries at ~2 MB RSS).

- Repo: https://github.com/tempoloss/quackiso
- Built and tested from a fresh `--recursive` clone against DuckDB v1.5.5; `LOAD` + `SELECT` returns correct rows.
- Rust extension using the C API (extension-template-rs pattern).

### DIFF
--- a/extensions/quackiso/description.yml
+++ b/extensions/quackiso/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: quackiso
-  description: Query ISO 20022 (camt/pacs/pain) financial messages as SQL
-  version: 0.0.2
+  description: Query ISO 20022 (camt, pacs) financial messages as SQL
+  version: 0.2.0
   language: Rust
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


tempoloss:
Re-pinned to v1.1.0 (`ff0b2cc466c1866b8a9437d3d0afd95f3a4dabc4`), which adds two table functions:

- `read_pacs004(path)` — pacs.004 payment returns, one row per `TxInf`. The returned amount is exposed beside the original settled amount, so a partial return is visible rather than looking like a full reversal.
- `read_pain002(path)` — pain.002 customer payment status reports, one row per status statement, with `status_level` naming which of the three nested levels a row came from. Only transaction rows carry an amount, so the coarser rows cannot double-count a total.

The description in this PR is updated accordingly (five functions, current corpus). Verified from a clean `--recursive` clone of the tagged ref in a Linux container: `make configure && make debug && make test` pass against `TARGET_DUCKDB_VERSION=v1.5.5`, and every statement in `hello_world` was run against the built extension.

`excluded_platforms` is unchanged. Happy to adjust or split this differently if you would rather review the original three functions first.

tempoloss:
Re-pinned to v1.2.0 (`13db4e434cb205378f29fafb0e701ebc7d5780e1`). Since the original submission the extension has grown from three table functions to thirteen — the full payment lifecycle (initiation, settlement, status, cancellation, resolution, returns and reversals, for both credit transfers and direct debits) — plus parallel multi-file scans (`threads := n`, one worker per file).

Verified from a clean `--recursive` clone of the tagged ref: `make configure && make debug && make test` pass against `TARGET_DUCKDB_VERSION=v1.5.5`, and every statement in `hello_world` runs against the built extension. `excluded_platforms` unchanged. Happy to split or trim if that makes review easier.

PR #2407, opened 2026-07-30
Add gdrive extension

jrosskopf:
Adds `gdrive`, a filesystem extension for **Google Drive** (not Google Cloud Storage).

It registers a `gdrive://` scheme, so anything that goes through DuckDB's virtual filesystem addresses a Drive file directly with no download or copy step:

```sql
CREATE SECRET gdrive (TYPE gdrive, PROVIDER service_account,
                      KEY_FILE '/etc/creds/sa.json', DRIVE_ID '0ABc...');

SELECT count(*) FROM 'gdrive://Finance/2026/actuals.parquet';
SELECT * FROM read_parquet('gdrive://exports/parts/*.parquet');
SELECT * FROM read_csv('gdrive://Finance/Budget');   -- a native Sheet
```

**What it covers:** ranged reads, glob, `COPY` writes, native Sheets/Docs via Drive's export, service-account and refresh-token auth, and DuckLake with its `DATA_PATH` on Drive.

**Testing.** There is no mock Drive and no HTTP replay layer — behaviour that needs a socket is tested against a real Google Drive account (live SQL suite plus a pytest harness), and pure logic is tested in Catch2. `v0.1.0` is green on DuckDB v1.5.5 and v1.4.5 LTS across Linux amd64/arm64, musl, Windows and both macOS architectures.

**Things reviewers usually ask:**

- *Excluded platforms* are `windows_amd64_mingw;wasm_mvp;wasm_eh;wasm_threads`. wasm is not a gap to close later: the extension is HTTPS calls to the Drive API plus a loopback redirect server for OAuth, neither of which works in a browser sandbox, and a wasm side-module would build green and then fail to load. A clean "unavailable" seemed better than an unloadable artifact.
- *Performance* is stated plainly in the description rather than sold. A cold 87 MB Parquet scan is ~3x the same file on GCS at the defaults and ~2x with a larger block size; a Drive ranged read costs ~1.3 s regardless of how few bytes it asks for. The numbers and the method are in [`docs/benchmark.md`](https://github.com/DataZooDE/duckdb-gdrive/blob/v0.1.0/docs/benchmark.md).
- *Known coverage gap:* rate-limit (403/429) handling is asserted against captured Drive response bodies rather than live, because provoking real quota exhaustion needs a throwaway Google project. It is documented in the README rather than left implied.

`ref` points at the `v0.1.0` tag commit.

### DIFF
--- a/extensions/gdrive/description.yml
+++ b/extensions/gdrive/description.yml
@@ -1,7 +1,6 @@
 extension:
   name: gdrive
   description: Query files in Google Drive directly with a gdrive:// filesystem
-  version: 0.1.0
   language: C++
   build: cmake
   license: MIT
@@ -17,7 +16,7 @@ extension:
 
 repo:
   github: DataZooDE/duckdb-gdrive
-  ref: 181b851d5297c5db743b9eaa8f978ae183580ff2
+  ref: c22f108fe83f849c0eb0bc5a271f54f14e1483b7
 
 docs:
   hello_world: |
--- a/extensions/gdrive/description.yml
+++ b/extensions/gdrive/description.yml
@@ -1,0 +1,95 @@
+extension:
+  name: gdrive
+  description: Query files in Google Drive directly with a gdrive:// filesystem
+  version: 0.1.0
+  language: C++
+  build: cmake
+  license: MIT
+  # wasm: the extension is HTTPS calls to the Drive API plus a loopback
+  # redirect server for interactive OAuth. Neither works in a browser sandbox,
+  # and a wasm side-module defers symbol resolution to load time -- so it
+  # would build green in CI and then fail to LOAD. A clean "unavailable" is
+  # better than an unloadable artifact.
+  # mingw: vcpkg's OpenSSL port is fragile there; sibling extensions skip it.
+  excluded_platforms: "windows_amd64_mingw;wasm_mvp;wasm_eh;wasm_threads"
+  maintainers:
+    - jrosskopf
+
+repo:
+  github: DataZooDE/duckdb-gdrive
+  ref: 181b851d5297c5db743b9eaa8f978ae183580ff2
+
+docs:
+  hello_world: |
+    -- Unattended: a service account. To WRITE you need a Shared Drive --
+    -- service accounts have no personal Drive storage quota and get
+    -- 403 storageQuotaExceeded anywhere else. Reading a shared folder is fine.
+    --
+    -- DRIVE_ID is a SHARED DRIVE id; ROOT_FOLDER_ID is a FOLDER id. They are
+    -- not interchangeable: passing a folder id as DRIVE_ID fails with
+    -- "Shared drive not found". Use whichever matches what was shared.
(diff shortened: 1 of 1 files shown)


jrosskopf:
Updated to **v2026.08.01** (`c22f108fe83f849c0eb0bc5a271f54f14e1483b7`), replacing the earlier v0.1.0 ref.

Since the original submission:

- **`PROVIDER credential_chain` is now the default**, so `CREATE SECRET (TYPE gdrive);` works with no arguments after a `gcloud` login. `PROVIDER authorization_code` adds browser consent with the user's own OAuth client.
- **A cold read across 150 files went from 17.3 s to 9.25 s**, with 298 fewer Drive API round trips, by not re-requesting metadata the extension had already received.
- Two correctness fixes: a dead file id was not recovered on the default read path, and a non-idempotent request could be retried on a 5xx.

I also dropped the `version:` field to match the other DataZoo descriptors — the project now uses CalVer, and `2026.08.01` is not valid semver, so the tagged `ref` is the version. Happy to put it back if you'd rather it were present.

Verified on the tagged commit: 937 pure assertions, 6 live SQL suites and 23 end-to-end tests against a real Google Drive account, plus CI green across Linux amd64/arm64, musl, Windows and both macOS architectures on DuckDB v1.5.5 and v1.4.5 LTS.

jrosskopf:
Retargeted to **v2026.08.03** (`3dc834f02c8934dc2a3dbafb664d2f4a62157a6e`), replacing the v2026.08.01 ref.

Twelve defects have been fixed since that tag, and **every one was a silent failure or a wrong answer rather than an error** — so none would be obvious to someone installing the current ref. The ones that matter most for a first-time user:

- **A missing secret was reported as a missing *file*.** `file_size` returned NULL and `remove_file` returned false when no secret was configured, so a typo in a secret name told you your data did not exist.
- **The extension could manufacture duplicate names on Drive.** Writing under a path whose parent was a file created a same-named folder beside it, leaving the original unreadable by path. Three separate code paths were affected; all are guarded now.
- **`glob` returned folders**, which callers then handed to `read_parquet`. Local `glob` of a directory returns zero rows; that parity is now matched.
- **`OpenFile` ignored `FILE_FLAGS_NULL_IF_NOT_EXISTS`**, which core relies on for speculative probes.
- Clearer errors for a revoked refresh token, an ADC credential with no quota project, and the `SCOPE` vs `DRIVE_SCOPE` mix-up (which previously reported "no secret configured" for a secret that plainly existed).

Two limitations are now documented rather than fixed, because Drive provides no primitive for them: concurrent writers can duplicate both files and directories, and `DRIVE_SCOPE` does not restrict an already-granted refresh token.

Verified on the tagged commit: 937 pure assertions, 7 live SQL suites and 23 end-to-end tests against a real Google Drive account, plus CI green across Linux amd64/arm64, musl, Windows and both macOS architectures on DuckDB v1.5.5 and v1.4.5 LTS.

Happy to squash the branch or rebase if that helps review.

PR #2411, opened 2026-07-30
laterite_ags4: bump to 0.9.0

niko86:
Updates the `laterite_ags4` community extension to 0.9.0, pinned to release commit 01ea351ba30e2ec1dfb75f834be2a07d287d40e8.


niko86:
Superseded by #2423, which bumps straight to 0.10.0.

The extension moved on after this PR's pin (`01ea351`): it now takes the `laterite-*` engine crates from crates.io rather than a git submodule, and laterite itself released 0.10.0 — which this extension's version tracks. Re-pointing this PR would leave its title and branch on the wrong version, so #2423 replaces it with a fresh per-version branch.

Nothing is lost: #2423 carries everything intended here plus the newer commits.

PR #2441, opened 2026-08-06
[Raster] Update extension to v1.2.0

ahuarte47:
Changes:
* New options in the `RT_CubeStats` and  `RT_CubeStats_Agg` functions to calculate statistics for a specific band (0-based index) in a set of datacubes, but only for those valid (non-nodata) cells that fall within a geometry (**Zonal statistics**).


ahuarte47:
Thanks @samansmink !!!!

PR #2464, opened 2026-08-08
[STAC]: Update extension to v0.2.0

ahuarte47:
Changes:
* Support for STAC API - Filter Extension (https://github.com/stac-api-extensions/filter).
* Support for STAC API - Fields Extension (https://github.com/stac-api-extensions/fields).
* Support for STAC API - Sort Extension (https://github.com/stac-api-extensions/sort).
* Catch error message when the STAC API endpoint returns an error with content_type as JSON.
* Fix management of STAC API endpoints that require one collection at least to be specified.
* Fix performance of requests caching with the default TTL to 30 seconds.
* Fix recursive reading of the STAC catalog.
* Remove dependencies of LibXml, servers do not send XML responses.

### DIFF
--- a/extensions/stac/description.yml
+++ b/extensions/stac/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: stac
   description: DuckDB extension for reading data from SpatioTemporal Asset Catalogs (STAC) using SQL.
-  version: 0.1.0
+  version: 0.2.0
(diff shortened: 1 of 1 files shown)


ahuarte47:
Thanks a lot @szarnyasg !

PR #2497, opened 2026-08-13
Add rocket

maxdemarzi:
Adds `rocket` — ROCKET (Dempster, Petitjean & Webb, 2020) random convolutional kernel features for time-series classification, as a scalar function.

```sql
SELECT rocket_transform(values, 250, 0, 0) AS features FROM series;
```

Convolves a series with random kernels of random length, dilation, padding and bias, reducing each kernel's response to two features — the global maximum and the proportion of positive values. Nothing is trained, so there is no model to fit or store.

### What is distinctive

Kernel `i` is a pure function of `(seed, i)`, derived through SplitMix64 used as a splittable seed source. So a caller can generate group `g` of a conceptual kernel bank without generating groups `0..g-1`:

```sql
SELECT rocket_transform(s, 4, 7, 4) = rocket_transform(s, 8, 7, 0)[9:16];  -- true
```

That is what makes it practical to split a 10,000-kernel bank into groups narrow enough for a downstream classifier's per-estimator feature budget, and it means kernels can be fanned across threads with no shared generator state.

Multivariate series (`DOUBLE[][]`, channels of timepoints) are supported; each kernel draws a random channel subset and still emits exactly two features, because the selected channels are summed inside one convolution. A one-channel series produces byte-identical kernels through either overload.

### Correctness

The pseudo-random stream is specified byte-for-byte in the project's `SPEC.md`, and the extension is validated against golden vectors emitted by an independent Python implementation: **max absolute difference 1.8e-15, PPV differences exactly 0**. The same conformance check runs on Linux/gcc and Windows/MSVC and reproduces to the same figure, so the spec is portable rather than accidentally shaped by one toolchain.

End to end, it has been run over a ten-dataset UCR subset composed with a tabular foundation model (`anofox_tabfm`), reproducing each dataset's reference accuracy exactly across two machines and two operating systems.

### Notes

- `excluded_platforms` omits the wasm targets.
- `ref` is pinned to a commit on `main`.

Thanks for maintaining this repository.

### DIFF
--- a/extensions/rocket/description.yml
+++ b/extensions/rocket/description.yml
@@ -1,0 +1,74 @@
+extension:
+  name: rocket
+  description: ROCKET random convolutional kernel features for time-series classification
+  version: 0.1.0
+  language: C++
+  build: cmake
+  license: MIT
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads"
+  maintainers:
+    - maxdemarzi
+
+repo:
+  github: maxdemarzi/duckdb-rocket
+  ref: e325474a7cab58bc11a30ebba61604002cb8fe55
+
+docs:
+  hello_world: |
+    -- 250 kernels produce 500 features: element 2i is kernel i's global max,
+    -- 2i+1 its proportion of positive values.
+    SELECT rocket_transform(values, 250, 0, 0) AS features FROM series;
+
+    -- Kernel i is a pure function of (seed, i), so a group of kernels is addressable
+    -- without generating the groups before it. These are the same kernels:
+    SELECT rocket_transform(s, 4, 7, 4) = rocket_transform(s, 8, 7, 0)[9:16];
+    -- true
+
+    -- Multivariate: outer list channels, inner list timepoints. Still 2 features
+    -- per kernel -- the selected channels are summed inside one convolution.
+    SELECT rocket_transform(channels, 250, 0, 0) FROM multivariate_series;
+
+    -- Variable-length: pass an explicit reference length so every row shares one
+    -- kernel bank. Without it each row draws its own, and column j stops meaning
+    -- the same thing across rows.
+    SELECT rocket_transform(values, 250, 0, 0, 64) FROM ragged_series;
+
+  extended_description: |
+    `rocket_transform` implements ROCKET (Dempster, Petitjean & Webb, 2020): it convolves a
+    series with random kernels of random length, dilation, padding and bias, and reduces each
+    kernel's response to two features — the global maximum and the proportion of positive
+    values. No training is involved, so there is no model to fit or store.
+
+    ```
+    rocket_transform(series DOUBLE[], kernels_per_group BIGINT, seed BIGINT, first_kernel BIGINT)
+      -> DOUBLE[]
+    ```
+
+    Features are interleaved: element `2i` is kernel `i`'s max and `2i+1` its PPV.
+
+    `first_kernel` selects a contiguous slice of one conceptual kernel bank. Kernel `i` is a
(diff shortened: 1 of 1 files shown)


maxdemarzi:
Dropped `excluded_platforms` — **wasm builds fine on all three targets**, so the exclusion was costing reach for no reason.

It had been in the descriptor since its first draft with no recorded justification. Rather than leave it, I ran the full distribution matrix on the extension's own CI at the same DuckDB version (v1.5.5, extension-ci-tools v1.5.5):

| platform | |
|---|---|
| linux_amd64, linux_arm64 | ✅ |
| osx_amd64, osx_arm64 | ✅ |
| windows_amd64, windows_amd64_mingw | ✅ |
| **wasm_mvp, wasm_eh, wasm_threads** | **✅** |

The extension is pure C++11 arithmetic over doubles with no dependencies, no threads, no filesystem and no network — the things a wasm exclusion normally exists for, none of which apply.

`ref` is unchanged: nothing under `src/`, `CMakeLists.txt`, `extension_config.cmake` or the duckdb submodule has moved, so the artifact built from that commit is identical.

jrosskopf:
Drive-by, since you have spent the day fixing bugs in [anofox_tabfm](https://github.com/DataZooDE/anofox-tabfm) and this looked like a quick one to hand back.

Both Linux builds fail here with:

```
make: *** No rule to make target 'set_duckdb_version'.  Stop.
```

That target comes from `extension-ci-tools/makefiles/duckdb_extension.Makefile`, which the extension's own `Makefile` is expected to `include`. At the pinned ref the repository root has neither:

| | `e325474` (pinned) | `main` |
|---|---|---|
| `Makefile` | absent | present |
| `extension-ci-tools` | absent (`.gitmodules` lists only `duckdb`) | present |

So the pin points at a commit from before the build scaffolding landed — `main` is 37 commits ahead of it, and carries both. Re-pinning `ref:` to a current `main` commit should be the whole fix.

For what it is worth, the addressable-kernel-group property is the part I found most interesting: being able to generate group `g` without generating `0..g-1` is exactly what makes the 10,000-kernel bank tractable against a model with a bounded feature budget — and it is the reason your `max_classes` and `max_features` observations over there were grounded in something real rather than hypothetical. Good luck with the review.

maxdemarzi:
Both failures came from the pinned `ref`, and I've pushed a fix.

**Root cause.** `ref: e325474a` predates this extension's build harness. At that commit there is no
`Makefile` and no `extension-ci-tools` submodule — the checkout log shows only `duckdb` being
registered:

```
git checkout --progress --force e325474a7cab58bc11a30ebba61604002cb8fe55
git submodule update --init --force --recursive
Submodule 'duckdb' (https://github.com/duckdb/duckdb.git) registered for path 'duckdb'
```

so `DUCKDB_GIT_VERSION=v1.5.5 make set_duckdb_version` had no makefile to read:

```
make: *** No rule to make target 'set_duckdb_version'.  Stop.
```

The `linux_amd64` job's `CANCELLED` is fail-fast collateral from the arm64 failure, not a separate
problem. That commit also predates the clang-format pass over `src/`, so `format;tidy` would very
likely have been the next failure even if the build had started.

**Fix.** Repointed at `c33d6b1a109e0ba9abf1722ad3fa424b79d31d86`, which has the `Makefile`, both
submodules, `vcpkg.json` and the formatted sources. That commit builds green across the full
platform matrix — including wasm — in the extension's own `MainDistributionPipeline`, using the same
`_extension_distribution.yml` this repo invokes.

Apologies for the noise; the ref was pinned before the CI files existed and I didn't re-pin it when
they landed.

PR #2500, opened 2026-08-14
quackiso 1.5.0: mandates, investigations, and payment status requests

tempoloss:
Re-pins `quackiso` to v1.5.0 (`af76fc5c236c5da1ea749b2860a6010b10b3be3f`). The registry currently builds v1.3.0 (`62c365d`), which predates twelve of the twenty-five readers. This PR opened as the 1.4.1 pin and now carries 1.5.0; the pacs.028 work below is unchanged and still part of it.

**New in 1.5.0: eleven readers, two families.**

`read_pain009` / `read_pain010` / `read_pain011` / `read_pain012` are the mandate's own lifecycle - the authorisation a direct debit pulls against, its amendment, its cancellation, and the report that accepts or refuses each of the three. A pain.008 collection already named a `mandate_id`; until now there was nothing to join it to. Grain is the record element, so a message stating three mandates is three rows, and the amendment carries both states: the mandate it changes and what it becomes.

`read_camt027` / `read_camt028` / `read_camt030` / `read_camt031` / `read_camt036` / `read_camt037` / `read_camt087` are the investigation family - the claim that money never arrived, the information that answers it, the notification that the case moved, the refusal, the debit authorisation request and its response, and the request to modify a payment rather than cancel it. Grain is the message. All seven expose the same assignment and case columns, so one case joins across every message that touches it, and camt.029 resolutions join to the claim they answer.

The assignment block is now parsed in one place (`wire::capture_assignment`) instead of three, and camt.029/055/056 call it. Their existing assertions are untouched, which is the proof the merge kept their behaviour.

Every column is proven non-NULL by a real fixture: twenty-three files from the ISO published business examples carried by `prog-nov/addon-iso20022` and `Nivaes/Iso20022Net`, provenance in each file header. Columns the schemas allow but no published sample carries were dropped rather than declared, and camt.037's first edition is read as it was spelled in 2005 - bare BICs, the payment stated inline - because it is the only sample carrying the amount the message exists to ask for.

**Descriptor changes.** `version` 1.5.0, `ref` on the tagged release commit, function list and corpus sentence updated to twenty-six functions and twenty-seven message families, and `hello_world` gains a `read_pain009` and a `read_camt027` statement.

**Previously in this PR, unchanged: `read_pacs028`** - payment status requests, the "where is my money?" message a bank sends when a payment it already submitted has gone quiet. A request carries no status of its own, so every monetary column on it is `original_*`. Two grains in one table, named by `scope`: a request detailing transactions gives one row per `TxInf`, and a request naming a whole original message with no transaction detail is still a row, because "where is my batch" is a question a bank does ask.

Also from the 1.4.0 audit and still included: a message's container is its scope instead of a latched flag, so a second message in one file cannot inherit the first one's identity; the counterparty is assembled as one party; CDATA is content on every path; an amount that fits `i128` but not `DECIMAL(38,5)` is refused instead of silently stored; February 31st is NULL instead of March 3rd; gzip is decided by the first two bytes rather than the file name; a directory matched by a glob is skipped instead of parsed.

**Verification, on the tagged ref `af76fc5`, all three workflows green:** Tests, which runs `make configure debug && make test` against `TARGET_DUCKDB_VERSION=v1.5.5` on Ubuntu plus the column coverage gate and the unit suite on Linux, macOS and Windows (https://github.com/tempoloss/quackiso/actions/runs/31884638263); Memory, the bounds run in release (https://github.com/tempoloss/quackiso/actions/runs/31884638287); Primitives, the doc anchors (https://github.com/tempoloss/quackiso/actions/runs/31884638274).

`cargo test` is 26 passed, 0 failed, 1 ignored, the ignored one being the 1.7 GB fixture Memory runs on a schedule. The SQL suite is 188 records. The coverage gate reports 25 readers, 61 files read, 472 columns, and it fails on any column no fixture proves non-NULL. Every statement in `hello_world` was run against the built extension with corpus paths substituted for the illustrative ones: eleven statements, eleven non-empty results.

### DIFF
--- a/extensions/quackiso/description.yml
+++ b/extensions/quackiso/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: quackiso
   description: Query ISO 20022 (camt, pacs, pain) financial messages as SQL
-  version: 1.4.1
+  version: 1.5.0
   language: Rust
   build: cargo
   license: MIT
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: tempoloss/quackiso
-  ref: 303af8767ebb3c299943201dad65739cb6614915
+  ref: af76fc5c236c5da1ea749b2860a6010b10b3be3f
 
 docs:
   hello_world: |
@@ -62,6 +62,19 @@ docs:
            original_settlement_date
     FROM read_pacs028('pacs028.xml');
 
+    -- The mandate itself (pain.009), which a direct debit pulls against:
+    -- who may collect, from whom, how often and up to what.
+    SELECT mandate_request_id, sequence_type, frequency, creditor_name,
+           debtor_name, collection_amount
+    FROM read_pain009('pain009.xml');
+
+    -- The investigation family (camt.027 here): the claim that money never
+    -- arrived. All seven investigation readers share the assignment and case
+    -- columns, so one case joins across the messages that answer it.
+    SELECT case_id, case_creator, assigner, assignee, original_amount,
+           original_settlement_date
+    FROM read_camt027('camt027.xml');
+
     -- Amounts are DECIMAL(38,5), so totals are exact.
     SELECT currency, SUM(amount) AS total
     FROM read_iso20022('statements/*.xml')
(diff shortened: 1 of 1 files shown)


tempoloss:
Windows was right to fail, and the failure was in the test rather than in the reader.

`test/sql/quackiso.test` globbed `testdata/camt053_*.xml` with `threads := 4` to assert that a malformed file fails the whole parallel scan. That glob holds two malformed fixtures, a bad amount and a truncated document, so the message the query carried was decided by whichever worker reached its error first. Linux in debug picked the bad amount, this job's Windows release build picked the truncated one, and the assertion named only the first. v1.3.0 passed the same job by the same luck.

Re-pinned to v1.4.1 (`303af8767ebb3c299943201dad65739cb6614915`), which asserts both errors against one malformed file per glob, five files each so the scan stays parallel. Six local runs of each glob give the same message every time. The built artifact is byte-identical to v1.4.0; the tag exists because the registry builds a ref.

PR #2503, opened 2026-08-15
gpudb: update to v0.4.0 (resident-column GPU functions; forward-looking cuda toolchain)

singhpratech:
Updates gpudb to v0.4.0 (ref `dc01848`).

**What's new in this version**

- A resident-column SQL surface: `gpu_upload(name, col)` uploads a column to GPU memory once, and `gpu_sum_resident` / `gpu_min_resident` / `gpu_max_resident` / `gpu_sum_resident_f64` then reduce it with zero per-query transfer. `gpu_last_stats()` reports which backend ran with kernel timings, and `gpu_build_info()` reports which backends the binary carries. Measured on TPC-H against native DuckDB v1.5.5 (results verified equal before timing): up to 25× on an RTX 4090 Laptop and 9.9× on an Apple M4 Max at SF100 — full grid with reproduction steps in the repo's BENCHMARK.md.
- The sqllogictest suite run by this repo's CI now covers the whole new surface (uploads, reductions, error cases, NULL semantics) with backend-independent assertions, so it passes on the GPU-less runners.
- Existing `gpu_sum/min/max` streaming aggregates are unchanged.

**About the `cuda` toolchain token — being upfront**

I've added `cuda` to `requires_toolchains` in a deliberately forward-looking way and want to be transparent about it rather than have it discovered in review:

- I'm aware the currently pinned `extension-ci-tools@v1.5-variegata` has no CUDA toolchain, so the token is inert today — the linux builds keep producing exactly the same CPU-fallback binary as v0.3.0 (the unknown token is ignored by the Docker toolchain switch; verified against the pinned branch).
- The CUDA toolchain support on extension-ci-tools `main` (the `cuda` case in `docker/linux_amd64`) is what this prepares for: once a release branch with it is adopted here, gpudb's build auto-detects nvcc and produces a CUDA-enabled linux binary with a **statically linked CUDA runtime** — no dynamic `libcuda`/`libcudart` dependencies, so `LOAD` keeps working on GPU-less machines with a clean CPU fallback. I validated that build end-to-end inside the `main`-branch linux_amd64 container with `extra_toolchains=";cuda;python3;"`: fatbins for sm_75–90, no CUDA libs in the ELF NEEDED entries, correct results with the driver present, clean fallback in a driver-less container.
- If you'd rather keep tokens strictly to what the pinned tooling supports, I'm happy to drop it from this PR and resubmit when the pin moves — whichever you prefer. As far as I can tell gpudb would be the first extension using the cuda toolchain, so I'd also gladly be a guinea pig for that rollout whenever it's useful to you.

Thanks as always for the review!

### DIFF
--- a/extensions/gpudb/description.yml
+++ b/extensions/gpudb/description.yml
@@ -1,89 +1,91 @@
 extension:
   name: gpudb
   description: GPU-accelerated analytical operators for DuckDB on NVIDIA CUDA and Apple Silicon Metal. First SQL execution engine that targets Apple Silicon GPUs.
-  version: 0.3.0
+  version: 0.4.0
   language: C++
   build: cmake
   license: Apache-2.0
-  requires_toolchains: "python3"
+  requires_toolchains: "python3;cuda"
   excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64;windows_amd64_rtools;windows_amd64_mingw"
   maintainers:
     - singhpratech
 
 repo:
   github: singhpratech/duckdbgpumetaldbram
-  ref: 2f416b1037a5d22662bd67b7fda26f0b96b82858
+  ref: dc01848708e8178c8cb07f551761f14002ffe931
 
 docs:
   hello_world: |
     LOAD gpudb;
-    -- Drop-in aggregates (BIGINT and DOUBLE overloads)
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #2504, opened 2026-08-15
Update UPDATING.md examples from quack to waddle

daviddallakyan2005:
## Summary

`UPDATING.md` still used `quack` as the C++ extension-template example. That name is now a core out-of-tree extension (`duckdb/duckdb-quack`); there is no `extensions/quack/` listing. The template listing is `waddle`.

This updates the workflow `extension_name` fields and the descriptor sample to match `extensions/waddle/description.yml`, and fixes the broken nested-backtick link (`[the quack extension](\`https://…\`)`) so it points at the live waddle descriptor.

## How to test

- Confirm `extensions/waddle/description.yml` exists and `extensions/quack/` does not.
- Open the descriptor link in the updated paragraph; it should resolve to `extensions/waddle/description.yml`.
- CI on this repo skips `UPDATING.md` (build workflows only watch `extensions/*/description.yml` and the build scripts), so no extension rebuild is expected.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -40,7 +40,7 @@ duckdb-stable-build:
   with:
     duckdb_version: v1.3.2
     ci_tools_version: v1.3.2
-    extension_name: quack
+    extension_name: waddle
 
 duckdb-next-build:
   name: Build extension binaries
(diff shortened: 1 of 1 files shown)


szarnyasg:
Thanks!

PR #2508, opened 2026-08-17
Add Data 360 extension

clawdbotborges:
## Summary

Adds `data360`, a read-only DuckDB extension for Salesforce Data 360 Query API V3. It provides bounded native HTTPS transport, Arrow IPC decoding, and interactive OAuth 2.0 Authorization Code authentication with S256 PKCE and a fixed loopback callback.

## Security model

- no client secret or refresh token
- bearer material remains in process memory
- DuckDB temporary secrets contain only opaque session references
- no Python, shell, broker, Salesforce CLI, or curl executable runtime dependency
- read-only Query API interface

## Platforms

Initial publication targets Linux x86_64, macOS arm64, and Windows x86_64/MSVC. Native-socket-incompatible Wasm, MinGW, Linux arm64, and macOS x86_64 targets remain excluded until separately proven.

## Verification

Local DuckDB v1.5.5 debug/release builds, 10/10 CTest, both SQLLogic suites, a 500-cycle callback-listener stress test, extension load verification, secret scanning, and prohibited-runtime scanning pass. The standalone source repository runs DuckDB’s official extension distribution workflow; this PR is draft while hosted matrix results complete.

### DIFF
--- a/extensions/data360/description.yml
+++ b/extensions/data360/description.yml
@@ -12,7 +12,7 @@ extension:
 
 repo:
   github: clawdbotborges/duckdb-data360
-  ref: 7127c8c3905675593f19b8fef58294141e4982a0
+  ref: fda061ad874a868c3f1177842bbe88a173e7136e
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Hi @clawdbotborges, could you have a look at the CI? There seems to be a linker issue.

PR #2509, opened 2026-08-17
[Raster] Update extension to v1.3.0

ahuarte47:
Changes:
* New optional `warp_options` parameter in the `RT_Read` and `RT_ReadCells` functions to reproject or warp the raster using GDAL Warp options (https://gdal.org/en/stable/programs/gdalwarp.html).
* Include PROJ database as an embedded binary stream, necessary for reprojection.
* Builds extension for `windows_amd64_mingw` arch as well.


ahuarte47:
Thanks a lot @szarnyasg !!

PR #2518, opened 2026-08-19
zim: v0.8.1

teaguesterling:
Updates `zim` from the registered **0.6.2** to **0.8.1**. Three releases have landed since, so this adds a capability rather than only bumping a version.

Pinned to the release commit `df80e1e` rather than a tag ref, so the build cannot be retargeted after the fact.

## v0.7.0 — the extension can now write archives

`COPY (query) TO 'out.zim' (FORMAT zim, …)` builds a ZIM from any relation, so the `description`, `hello_world` and `extended_description` are updated — the entry previously described a read-only extension.

Columns resolve by name (`path` and `content` required; `title`, `mimetype`, `is_redirect`/`redirect_path` and libzim's hints optional), and `read_zim`'s own output schema is accepted directly, so a ZIM→ZIM copy is one statement. Metadata, a cover illustration, the main page, compression and a Xapian index are set through copy options, so a written archive is immediately searchable with `zim_search`.

Writing never overwrites an existing file, a failed copy leaves nothing behind, and `MAIN_PATH` and every redirect target are validated against the entries actually written — libzim silently accepts a missing main page and silently drops dangling redirects.

## v0.8.0 — pushdown correctness, cancellable scans, streaming `zim://`

- A pushed-down `WHERE mimetype …` was OR-ed into the `mimetype :=` argument instead of intersected, so a `WHERE` clause could make a query return **more** rows than the argument allowed.
- Long scans with a selective predicate could not be interrupted; measured 8.5s → 1.0s to respond to `Ctrl-C` on a 3M-entry archive.
- The `zim://` filesystem serves ranged reads instead of materialising whole entries — a 268 MB entry drops from 557 MB to 34 MB peak for a ranged consumer, and whole-entry reads got *faster* (48.7 ms → 25.7 ms). Also fixed an out-of-bounds read where `location + nr_bytes` could overflow past the bounds check.

## v0.8.1 — replaced archives no longer serve stale bytes

The archive pool cached an opened handle keyed by path and never revalidated it, so replacing a `.zim` in place — which is how Kiwix ships updates — kept serving the old archive for the life of the connection. It now reopens when `(dev, inode, mtime, size)` changes. Remote archives are exempt, since there is no `stat` over HTTP.

## Build metadata

`vcpkg_commit` is unchanged and still matches the extension's own CI pin. `excluded_platforms` is unchanged (Windows is still unsupported).

CI green on Linux x64/arm64, macOS x64/arm64, all three WASM variants, ThreadSanitizer, and the remote-search integration test. 538 assertions plus four shell-level harnesses.

### DIFF
--- a/extensions/zim/description.yml
+++ b/extensions/zim/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: zim
-  description: Read .zim (Kiwix / openZIM) archives directly in DuckDB via libzim, from local files or remote S3/HTTP — offline Wikipedia, WikiMed, Stack Exchange, iFixit, and more, with a zim:// filesystem and full-text search.
-  version: 0.6.2
+  description: Read and write .zim (Kiwix / openZIM) archives directly in DuckDB via libzim, from local files or remote S3/HTTP — offline Wikipedia, WikiMed, Stack Exchange, iFixit, and more, with a zim:// filesystem, full-text search, and COPY TO for building archives from any query.
+  version: 0.8.1
   language: C++
   build: cmake
   license: GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

PR #2522, opened 2026-08-19
[Raster] Update extension to v1.5.0

ahuarte47:
Changes:
* New `RT_Create` function to create a new raster file with specified parameters (CRS, BBOX, raster size, number of bands, data type, nodata value, etc.).
* Add `lerc`, `webp` and `zstd` encoding to Gtiff reader.
* New `ignore_nodata` parameter in the `RT_ReadCells` function. It allows to ignore cells with nodata values when reading raster data. The `ignore_nodata` parameter can take the following values:
  - `0`: Never ignore cells (default).
  - `1`: Ignore cells with at least one band having a nodata value.
  - `2`: Ignore cells with all bands having nodata values.

### DIFF
--- a/extensions/raster/description.yml
+++ b/extensions/raster/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: raster
   description: DuckDB extension for reading and writing geospatial raster data using SQL.
-  version: 1.3.0
+  version: 1.4.0
   language: C++
   build: cmake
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

ahuarte47:
Thanks a lot @sebastiaan-dev !!!

PR #2524, opened 2026-08-20
Add gridpin_ext: offline geocoding from a single country file

gridpin-reserve:
Adds `gridpin_ext` — offline forward and reverse geocoding against a single memory-mapped
country file. No server, no network calls, no per-request fees: the addresses being geocoded
never leave the machine.

### Source

https://github.com/gridpin/duckdb-gridpin at `cf94e5d3994e48fbf12b67a2b808cdde9bda428d`
(signed commit). The repository root is the extension crate in the layout
`extension-template-rs` expects. The geocoding engine lives in a separate repository and is
pulled in as a git dependency pinned to an exact commit, so a build here is reproducible.

### Build status

The extension repository's own `Main Extension Distribution Pipeline` is green on this exact
commit — `linux_amd64`, `linux_arm64`, `osx_amd64`, `osx_arm64`, `windows_amd64` and
`windows_amd64_mingw` all build and pass the SQL tests:

https://github.com/gridpin/duckdb-gridpin/actions/runs/32357704052

`excluded_platforms` lists only what we have not built: the wasm targets (the engine
memory-maps a local file), `linux_amd64_musl` and `windows_amd64_rtools`.

### Functions

| Function | Purpose |
|---|---|
| `gridpin_load(path)` | memory-map a country sheet |
| `gridpin_geocode(address)` | address string → JSON with `lat`, `lon`, `street`, `city` |
| `gridpin_reverse(lat, lon)` | coordinates → JSON for the nearest address |
| `gridpin_load_poi(path)` | optional points-of-interest layer |
| `gridpin_reset()` | drop the loaded sheet so another country can be loaded |

A query issued before `gridpin_load` is a clear error rather than a crash, and an address with
no match returns `{}` — an empty result is a documented contract, not a guess.

### Data and licensing

The extension and the engine are Apache-2.0. Country data files are a separate free download
(https://gridpin.dev): France comes from the national BAN registry under Licence Ouverte;
Italy, the Netherlands and Serbia are built from Overture. Each sheet carries its own source
licence and attribution.

### DIFF
--- a/extensions/gridpin_ext/description.yml
+++ b/extensions/gridpin_ext/description.yml
@@ -1,0 +1,65 @@
+extension:
+  name: gridpin_ext
+  description: Offline geocoding from a single country file — forward and reverse, no server, no network
+  version: 0.1.0
+  language: Rust
+  build: cargo
+  license: Apache-2.0
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;linux_amd64_musl"
+  requires_toolchains: "rust;python3"
+  maintainers:
+    - gridpin-reserve
+
+repo:
+  github: gridpin/duckdb-gridpin
+  ref: cf94e5d3994e48fbf12b67a2b808cdde9bda428d
+
+docs:
+  hello_world: |
+    -- Country data files are free downloads: https://gridpin.dev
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

PR #2532, opened 2026-08-22
Update dicom extension to v0.7.0

nmontesg:
This PR updates the `dicom` extension to version v0.7.0:
* The main change is a fixing of a memory leak in the `retrieve_dicom` function when using `load_pixel_data=false` (the default arg value)
* Other minor changes include better includes organization, documentation fixes for the networking functions, and registering function information objects (instead of functions) to expose their documentation in table `duckdb_functions()`.


sebastiaan-dev:
Thanks!

PR #2539, opened 2026-08-23
Add fusion_scanner 

krokozyab:
Query Oracle Fusion from DuckDB through BI Publisher, with SSO, an ATTACH-able catalog and cached metadata


sebastiaan-dev:
Thanks!

PR #134, opened 2024-10-08
Add `sheetreader` as community extension

freddie-freeloader:
Hi!

In the last semester, I was part of a programming project organized by the DIMA group at TU Berlin. We created a small DuckDB-extension named [`sheetreader`](https://github.com/polydbms/sheetreader-duckdb) that utilizes [sheetreader-core](https://github.com/polydbms/sheetreader-core) (a fast multi-threaded XLSX parser) for importing XLSX files into DuckDB.

We did a few benchmarks comparing our extension to the import function which the `spatial` extension provides (`st_read`). Our first benchmarks indicate, that depending on several factors the `sheetreader` extension is around 5 to 10 times faster than the `spatial` extension at parsing XLSX files and loading them into DuckDB (https://github.com/polydbms/sheetreader-duckdb/?tab=readme-ov-file#benchmarks).

We would like to offer this extension as a DuckDB community extension.

A note regarding the repository structure of our extension:
- We have a version in branch [`benchmark-version`](https://github.com/polydbms/sheetreader-duckdb/tree/benchmark-version) that has code dedicated for benchmarking.
- On the branch [`main`](https://github.com/polydbms/sheetreader-duckdb/tree/main), we provide a “slimmed down” version with that code removed. We would like to offer the latter version as the community extension.

Let me know if there are adjustments to be made. :slightly_smiling_face:

PR #2542, opened 2026-08-24
webbed: v2.7.0

teaguesterling:
Bumps `webbed` to v2.7.0 — a correctness release covering five reported issues.

- **#129** — `html_extract_*` now accepts plain `VARCHAR`, so `html_extract_text(content, '//h2')` works without an explicit `::HTML` cast.
- **#130** — `html_extract_tables_json` was unusable: a table with no `<th>` raised an `INTERNAL` error, and the declared return type never matched the value (the `rows` element type was derived from the document's own header names, so any cast or field access failed). The content-dependent field is replaced by a fixed-shape `table_json VARCHAR`.
- **#131** — `<tfoot>` rows were dropped or silently relabelled as data; `html_extract_table_rows` now reports `row_type = 'footer'`.
- **#132** — control characters below `0x20` were copied through verbatim, producing invalid JSON.
- **#134** — a syntactically invalid XPath silently returned empty, indistinguishable from a valid expression that matched nothing. It now raises. An *undeclared namespace prefix* still returns empty, since that fails at evaluation rather than at parse.

Release notes: https://github.com/teaguesterling/duckdb_webbed/releases/tag/v2.7.0

`ref` is the raw v2.7.0 commit (`c1b32ef`) rather than a tag ref. All shippable platforms are green on that commit: linux amd64/arm64, osx amd64/arm64, windows amd64, and wasm mvp/eh/threads.

`andium` (DuckDB v1.4.5 track) stays at its prior commit; these fixes ship on the v1.5.x track via `ref`.

### DIFF
--- a/extensions/webbed/description.yml
+++ b/extensions/webbed/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: webbed
   description: Comprehensive processing extension for web markup languages (XML and HTML) with SAX streaming for large files, intelligent schema inference, XPath-based data extraction, and HTML table parsing.
-  version: 2.6.1
+  version: 2.7.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

PR #2544, opened 2026-08-25
quackiso 1.7.0: an address audit over both wire formats, MT101 and MT104

tempoloss:
Three new functions, and the reason for all three is the CBPR+ deadline.

On 14 November 2026 CBPR+ stops accepting a fully unstructured postal address, with no grace period. `audit_addresses(path)` answers the question that comes before that migration: of the traffic already on disk, which parties would be refused and why. One row per party occurrence, with the shape read off the wire - STRUCTURED, HYBRID, UNSTRUCTURED or NONE - and a `finding` naming what the rule would refuse.

It reads SWIFT MT as well as ISO 20022, and that is the point rather than a bonus. An MT `:50K:` is a name and then free-text address lines, which is exactly the shape that stops being accepted, so a migration answered on the XML half only is answered on the easier half. One classifier grades both wires, so an MT party and an ISO 20022 party cannot disagree about what UNSTRUCTURED means.

The audit reports the address lines themselves beside the counts, because a refusal names what is missing while the lines are what there is to work with, and neither follows from the other. A free-text address whose town is already written needs its town moved into an element; a bare one needs a town collected. Both carry the same finding. Deciding which is which is a reading of the data, so the function reports the lines and stops there rather than guessing a town out of free text.

`read_mt101(path)` and `read_mt104(path)` are the request for transfer and the direct debit: the FIN originals of pain.001 and pain.008, the initiation and collection halves that this extension could read on the ISO 20022 side and on neither MT side. Each repeats a transaction sequence, so the grain is the transaction and the header repeats on every row. An MT104 also states a batch total after its last transaction, reported beside the sum of the transactions because the format writes `:19:` only when the two differ.

Thirty-seven functions now: thirty-five readers, the sniffer, and the audit. Tested against the same real corpora as before, now including generated MT101 and MT104 messages from the swift-mt-message scenario set.

### DIFF
--- a/extensions/quackiso/description.yml
+++ b/extensions/quackiso/description.yml
@@ -1,7 +1,10 @@
+# Submit this file to duckdb/community-extensions at
+# extensions/quackiso/description.yml
+# `ref` is the tagged release commit in tempoloss/quackiso.
 extension:
   name: quackiso
   description: Query ISO 20022 (camt, pacs, pain) and SWIFT MT financial messages as SQL
-  version: 1.6.2
+  version: 1.7.0
   language: Rust
   build: cargo
   license: MIT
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

PR #2546, opened 2026-08-25
Add duckdb_rphonetic: Kölner Phonetik and Daitch-Mokotoff Soundex

guizmaii:
Adds `duckdb_rphonetic`, two phonetic encoders for European name matching.

| Function | Signature |
|---|---|
| `cologne_phonetic` | `VARCHAR -> VARCHAR` |
| `daitch_mokotoff` | `VARCHAR -> VARCHAR[]` |

```sql
SELECT cologne_phonetic("Müller"), cologne_phonetic("Mueller");   -- 657, 657
SELECT list_has_any(daitch_mokotoff("Klein"), daitch_mokotoff("Cleyn"));  -- true
```

**Why:** core DuckDB ships `soundex` and `double_metaphone`, both tuned for English. Neither Kölner Phonetik (the German-language counterpart to Soundex) nor Daitch-Mokotoff (built for Germanic and Slavic surnames) is available. Trigram similarity is not a substitute — `Klein` and `Cleyn` share no trigrams at all, yet both encoders match them.

Daitch-Mokotoff returns a list rather than a scalar because ambiguous spellings legitimately encode several ways; matching is a set overlap via `list_has_any`.

**Implementation:** a thin wrapper over the [`rphonetic`](https://crates.io/crates/rphonetic) crate by Dalvany, a Rust port of Apache Commons Codec. Apache-2.0, with a `NOTICE` crediting both.

**Testing:** Apache Commons Codec 1.22.1 is used as an oracle over a 279-name corpus. `test/oracle/run.sh` regenerates the expected values from the Maven Central jar, so the committed fixtures are reproducible rather than asserted. Daitch-Mokotoff agrees with Commons Codec on all 279 names, values and order.

**One behavioural note, documented rather than hidden.** Kölner Phonetik implementations disagree on whether a dropped `0` breaks a run of identical digits. This extension follows `rphonetic` — `cologne_phonetic("Dieter") = "227"` where Commons Codec gives `27` — and the two differ on 36 of the 279 names. Every divergence is committed and pinned by a test, so an upstream bump cannot change output silently. The published reference vectors (`Müller-Lüdenscheidt` → `65752682`, `Wikipedia` → `3412`, `Breschnew` → `17863`) are unaffected.

CI is green on all 9 platforms at the pinned ref (linux amd64/arm64, macOS x86_64/arm64, Windows amd64/mingw, wasm mvp/eh/threads), with the SQL suite executing on linux_amd64, both Windows targets and osx_arm64.

### DIFF
--- a/extensions/duckdb_rphonetic/description.yml
+++ b/extensions/duckdb_rphonetic/description.yml
@@ -5,6 +5,8 @@ extension:
   language: Rust
   build: cmake
   license: Apache-2.0
+  requires_toolchains: "rust;python3"
+  excluded_platforms: "linux_amd64_musl"
   maintainers:
     - guizmaii
 
--- a/extensions/duckdb_rphonetic/description.yml
+++ b/extensions/duckdb_rphonetic/description.yml
@@ -1,0 +1,53 @@
+extension:
+  name: duckdb_rphonetic
+  description: Phonetic algorithms for German and European name matching
+  version: 0.1.0
+  language: Rust
+  build: cmake
+  license: Apache-2.0
+  maintainers:
+    - guizmaii
+
+repo:
+  github: guizmaii-opensource/duckdb-rphonetic
+  ref: a7cee7a7ab8f7d32fd1474b66b9ba6ac2f6f27d2
+
+docs:
+  hello_world: |
+    -- Kölner Phonetik: a German-language counterpart to Soundex.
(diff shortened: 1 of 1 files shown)


guizmaii:
Ready for review

sebastiaan-dev:
Thanks!

PR #2547, opened 2026-08-25
oracle_scanner: bump to 0.2.0

krokozyab:
Opens an OCI wallet through its auto-login cwallet.sso, so WALLET_PASSWORD is no longer required. The ref is the commit tagged v0.2.0


sebastiaan-dev:
Thanks!

PR #2563, opened 2026-08-28
gpudb: update to v0.6.0

singhpratech:
Updates `gpudb` from v0.5.0 to v0.6.0 (`ref` → `5ddcb0fa2d74a67a72a5197f5848cdc3bd84ee66`, tag `v0.6.0`).

What's new in this version:
- Resident `GROUP BY` from SQL on both backends — `gpu_groupby_{sum,sum_f64,count}_resident(name)` return `(key, aggregate, count)` rows over an uploaded `(key, payload)` pair, using the same device-cached sort the joins use as a build side.
- `HAVING` and `ORDER BY aggregate LIMIT k` evaluated on the device — the `_having(name, cmp, threshold)` and `_topk(name, k, order)` forms return only the surviving groups.
- `gpu_topk_resident` (`ORDER BY col LIMIT k` over a resident column).
- Fixes: a Metal radix-sort pass-skip bug that could mis-sort certain key sets (affected the v0.5.0 Metal join build cache for those keys), Metal device memory now released on drop (ARC), CUDA device faults surface as SQL errors instead of aborting.

Results are verified equal to native `GROUP BY` both ways on TPC-H SF1/10/50 on both backends; numbers and their caveats are in the repo's BENCHMARK.md and in the extended description here. The community build path (`make configure && make release && make test`) was run on the tagged commit on Linux with the CUDA toolchain and on macOS.

Fork validation run of this branch (same workflow, all four platforms green): https://github.com/singhpratech/community-extensions/actions/runs/33217828943

Thanks for the sweep as always.

### DIFF
--- a/extensions/gpudb/description.yml
+++ b/extensions/gpudb/description.yml
@@ -130,7 +130,10 @@ docs:
     Source: https://github.com/singhpratech/duckdbgpumetaldbram
 
   hardware_requirements: |
-    NVIDIA CUDA: any GPU with sm_75 or later. Driver 525+ recommended.
+    NVIDIA CUDA (release binary or source build, not the community binary):
+    any GPU with sm_75 or later; the release binary's static CUDA 13 runtime
+    needs driver 580 or newer, a source build needs only the driver matching
+    your toolkit.
     Apple Metal: any Apple Silicon Mac (M1 or later). macOS 13+ for
     int64 atomics; macOS 15+ for MSL 3.2.
     Falls back to CPU otherwise.
--- a/extensions/gpudb/description.yml
+++ b/extensions/gpudb/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: gpudb
-  description: GPU-accelerated aggregates and joins for DuckDB on NVIDIA CUDA and Apple Silicon Metal — upload columns once, then reductions and joins run on the GPU. First SQL execution engine that targets Apple Silicon GPUs.
-  version: 0.5.0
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

singhpratech:
Thank you @sebastiaan-dev — and for getting to it on a Saturday, no less. Just confirmed `FORCE INSTALL gpudb FROM community` on v1.5.5 comes up as v0.6.0 with the new GROUP BY / HAVING / top-k functions working. Much appreciated, as always.

singhpratech:
Hi @sebastiaan-dev — sorry to bother you again, and thank you once more for the quick merge on Saturday, it's very much appreciated.

One small question, purely in case it's something on our side: the registry itself has been serving gpudb 0.6.0 since the merge (`INSTALL gpudb FROM community` on v1.5.5 confirms `extension_version = v0.6.0`), but the docs page at https://duckdb.org/community_extensions/extensions/gpudb.html still shows the 0.5.0 description. Is the page refresh simply on its own schedule (in which case, apologies for the noise — happy to just wait), or is there anything we should have updated in our `description.yml` to trigger it?

No urgency at all. Thanks again for all the work you folks put into the registry — it shows.

sebastiaan-dev:
@singhpratech it seems some extensions have failed deploying, I will look into this and see if this is the case for you as well.

sebastiaan-dev:
@singhpratech your extension is good to go, the docs update takes a bit longer but you can see your changes here: https://github.com/duckdb/duckdb-web/pull/7190.

singhpratech:
@sebastiaan-dev - you are amazing ! thanks again for being so responsive and helpful !!

PR #2566, opened 2026-08-29
fusion_scanner 0.3.0

krokozyab:
The attached catalog now lists every table in the dictionary, and cached metadata no longer expires by age. Pins ref to 8490c81, whose full build matrix passed on DuckDB v1.5.5.


sebastiaan-dev:
Thanks!

PR #2574, opened 2026-08-29
duck_hunt: update to v1.11.1

teaguesterling:
Updates `duck_hunt` to version **v1.11.1**.

### Changes
- Fix structured_data JSON escaping across 25+ parsers (#55)
- Replace regex named-group extraction with zero-regex string scanner to eliminate MSVC complexity crashes
- Add exception containment across regex auto-detect and command lookup
- Fix GTest singular vs plural pattern matching and suite end delimiter evaluation order


sebastiaan-dev:
Thanks!

PR #2582, opened 2026-08-31
add quackhole: connect to duckdb behind NAT via iroh

smithclay:
New extension, somewhat building on the previous [quackscale](https://github.com/Query-farm/quackscale) extension but using a different protocol, [iroh](https://docs.iroh.computer/)

Enables several cool things, reason for writing it was making it easy to get a p2p connection to a duckdb instance, including from inside a browser, without needing to define tailscale networks, overlays, certificates, etc. Kind of a "duckdb meets ngrok" sort of deal.

https://github.com/smithclay/quackhole


lghavami:
Hi @smithclay all looks good! Thanks :)

PR #2587, opened 2026-09-01
Add ossie

venkata-chikkam:
Adds `ossie` — the Apache Ossie (incubating) reference implementation for DuckDB.

Apache Ossie is a vendor-neutral file format for semantic models: datasets bound to physical tables,
fields, declared relationships, and metrics written as aggregate expressions. It describes data in
place and never transforms it. What the format deliberately leaves open is the *query* those
definitions go into — the joins, the `GROUP BY`, and the grain — because those depend on the question
being asked.

Every other shipped Ossie implementation is a converter that hands those definitions to some vendor's
semantic layer. This extension executes the query itself: it plans the joins, derives the grain from
the declared keys, and emits SQL DuckDB runs. It reads both YAML and JSON models, and
`examples/server.sql` publishes a model over MCP so an agent can discover the vocabulary and ask
questions against it.

```sql
INSTALL ossie FROM community;
LOAD ossie;
CALL ossie_load('model.yaml');
SELECT * FROM ossie_query(['revenue'], ['customers.country']);
```

### Notes for review

- **No runtime dependencies.** JSON parsing uses the yyjson DuckDB already vendors; YAML uses fkYAML,
  pinned as a submodule. No vcpkg, no `install_notes`, nothing for a user to install separately.
- **`extension_config.cmake` also loads `tpcds`.** It is a test dependency only — nothing in `src/`
  references it. The differential tests generate TPC-DS data with `dsdgen` and diff this extension's
  generated SQL against hand-written TPC-DS SQL, which is the only layer that catches a wrong
  *number* rather than wrong-looking SQL. It is not linked into the shipped extension.
- **Adds no settings and no types**, and registers seven functions, each with a description and
  runnable examples so the generated documentation page is populated.
- **Built and tested against this exact ref** on a fork of this repo before opening:
  [run 33475438150](https://github.com/iqea-ai/community-extensions/actions/runs/33475438150) — all
  ten platform jobs green, and the build ran the extension's own suite (210 assertions in 12 test
  cases), including the tests that depend on `tpcds`.
- **Conformance.** Models are validated against the format's own `core-spec/osi-schema.json`, and the
  test suite includes five models published by other Ossie implementers (Databricks, GoodData,
  NVIDIA, Omni, OrionBelt) vendored verbatim from `apache/ossie`. Four load and answer queries; the
  fifth carries only `DATABRICKS` expressions, which this extension does not execute.
- **Naming.** This is an independent implementation of the Apache Ossie file format, stated in the
  README, the `NOTICE`, and the extended description. It is not affiliated with or endorsed by the
  ASF.

### DIFF
--- a/extensions/ossie/description.yml
+++ b/extensions/ossie/description.yml
@@ -8,6 +8,7 @@ extension:
   build: cmake
   license: MIT
   maintainers:
+    - venkata-chikkam
     - iqea-ai
 
 repo:
--- a/extensions/ossie/description.yml
+++ b/extensions/ossie/description.yml
@@ -1,0 +1,112 @@
+extension:
+  name: ossie
+  description: The Apache Ossie (incubating) reference implementation for DuckDB. Reads Ossie
+    semantic model files in YAML or JSON and answers semantic queries against the tables in DuckDB,
+    from a single binary, with no infrastructure. Also exposes the semantic layer via MCP.
+  version: 0.1.0
+  language: C++
+  build: cmake
+  license: MIT
+  maintainers:
+    - iqea-ai
+
+repo:
+  github: iqea-ai/duckdb-ossie
+  ref: ef995aa581455840c7adc4e24d30220b047962a3
+
+docs:
+  hello_world: |
+    INSTALL ossie FROM community;
+    LOAD ossie;
+
+    -- A model is a file you supply. This one is written inline so the example runs as pasted;
+    -- normally you would point ossie_load at a .yaml or .json file from your own repository.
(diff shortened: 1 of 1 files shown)


lghavami:
Hi @venkata-chikkam, thanks for the PR! I see that @iqea-ai is the owner of the repo and listed as the maintainer; therefore they should make a PR themselves and/or authorize you as a maintainer (also via a PR).

venkata-chikkam:
Thanks, good catch. iqea-ai is the organization that owns the repo, so it can't open a PR itself. I've added venkata-chikkam to maintainers in this same PR, which should cover the authorization. My Git profile also should reflect the same.

lghavami:
perfect, everything looks good!

PR #2588, opened 2026-09-01
Bump gdrive to v2026.09.01

jrosskopf:
Updates `gdrive` to [`v2026.09.01`](https://github.com/DataZooDE/duckdb-gdrive/releases/tag/v2026.09.01).

## What changed

**Workload identity federation for `credential_chain`.** `external_account` credentials were previously recognised and refused, which meant the extension could not be used from a cluster without a downloaded service-account key to store, mount and rotate. It now performs the exchange directly (STS token exchange, then service-account impersonation with the Drive scope), consistent with how the existing RFC 7523 path is implemented rather than pulling in `google-cloud-cpp`.

Only file-sourced subject tokens are supported — a projected Kubernetes token, for example. `url`-sourced tokens (AWS/Azure/GCE metadata) keep the previous "recognised but unsupported" behaviour with a named error, rather than being accepted and failing later.

The key-file (`PROVIDER service_account`) and interactive (`gcloud auth application-default login`) paths are unchanged.

## Docs

`extended_description` gains a short paragraph on this. The published docs previously showed only the key-file and interactive paths, which is exactly what this release makes unnecessary for a server.

## Verification

Tested against real Google Drive, no test doubles — the upstream repo builds against live Drive by policy:

- A real projected Kubernetes token → real STS → real service-account impersonation → listing a real Shared Drive
- Pure-logic parsing covered by Catch2 (944 assertions, 195 cases); the network exchange is covered live, not mocked
- `gdrive_version()` and the release tag verified in agreement by the repo's stamp check

Full build matrix green on both DuckDB lines at the tagged commit.

### DIFF
--- a/extensions/gdrive/description.yml
+++ b/extensions/gdrive/description.yml
@@ -16,7 +16,7 @@ extension:
 
 repo:
   github: DataZooDE/duckdb-gdrive
-  ref: e4d78a732f38225a9690e99828cc5c5e72a8a8d1
+  ref: 73b8a5e729f29338d883b5277b99644be66c8c4b
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


lghavami:
All looks good @jrosskopf, thanks!

PR #2589, opened 2026-09-01
orc: bump to v0.2.0 — write support (COPY ... TO FORMAT orc)

alitrack:
## Summary

Bump the `orc` extension (`alitrack/duckdb_orc`) from ref `1dab6ff` (v0.1.3) to `891ecbd` (**v0.2.0**).

### v0.1.3 (previous)
- **feat: stripe-level parallel ORC reading** (`367e51f`) — `read_orc` reads ORC stripes in parallel across threads (caps at CPU count).

### v0.2.0 (this bump) — **Write support**
- **`COPY tbl TO 'file.orc' (FORMAT orc)`** — write any query result to ORC via the DuckDB C API COPY function interface (bind → global_init → sink → finalize)
- **ZSTD compression** by default (requires `orc-rust` ≥ `2060e243`, where compressed writes landed upstream 2026-08-18)
- Supported types: bool, int8/16/32/64, float32/64, date, timestamp, varchar, blob
- Unsupported types (List/Struct/Map/Decimal) rejected with a clean binder error — upstream `orc-rust` writer would otherwise panic and abort the process
- Custom C entrypoint (the `duckdb_entrypoint_c_api` macro doesn't expose the raw `duckdb_connection` needed to register a COPY function)
- Roundtrip integration tests: write → read back with count/sum/date/timestamp assertions

### Known limitations
- Output columns are named `column_0, column_1, ...` — the DuckDB C API's COPY bind does not expose source column names
- Writer type coverage limited to upstream `orc-rust` writer support (no List/Struct/Map/Decimal yet)

### DIFF
--- a/extensions/orc/description.yml
+++ b/extensions/orc/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: orc
   description: A DuckDB extension for reading Apache ORC files, written in pure Rust. Supports all ORC primitives, struct, list, map types and all compression codecs (Zlib, Snappy, LZO, LZ4, ZSTD).
-  version: 0.1.3
+  version: 0.2.0
   language: Rust
   build: cargo
   license: MIT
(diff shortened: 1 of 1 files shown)


lghavami:
@alitrack all merged, thanks!

PR #2590, opened 2026-09-01
Update `textplot` to latest commits

rustyconover:


lghavami:
Hi @rustyconover all looks good! Given that you have another PR open for the same extension I will go ahead and close that one. Thanks :)

PR #2592, opened 2026-09-02
Update anofox_forecast to a9328e1 (v0.9.0)

sipemu:
Bumps the pinned ref to the v0.9.0 release (adds the DuckDB-Wasm runtime harness + CI gating and the ensemble surface).

Also excludes windows_amd64: the community build_all LTS leg uses ci_tools_version v1.4-andium, whose vcpkg baseline pins an OpenSSL dependency with a now-404 upstream download ('Download failed, halting portfile'). Excluding windows_amd64 keeps the deploy pipeline green; the v1.5.x windows_amd64 build works and is unaffected upstream.


lghavami:
Thanks!

PR #2594, opened 2026-09-02
trino_parity: bump to 0.3.0 (simple per-code-point case mapping, matches Trino)

apatrida:
Bumps `trino_parity` to `b181c61c165b9f79e7902ee73de509f14b391299` (version 0.3.0).

### What changed

`trino_lower` / `trino_upper` now apply **simple per-code-point** Unicode case
mapping (ICU `u_tolower` / `u_toupper`). The previous release used ICU full
string case mapping (`UnicodeString::toLower/toUpper(Locale::getRoot())`),
which was the wrong model for Trino: Trino's `lower()`/`upper()` go through
airlift `SliceUtf8` → `Character.toLowerCase(int)` / `toUpperCase(int)`, so
`upper('ß')` is `'ß'` (not `'SS'`), `lower('İ')` is `'i'` (not `'i'` + U+0307),
there is no Greek final-sigma rule, and ligatures are unchanged. Verified
against Trino 483 on a 63-string corpus: 16 mismatches before, 0 after
(one residual from Unicode-version skew in the vendored ICU, tracked upstream).

Everything else (`trino_reverse`, trim family, `trino_normalize`, the three
hashes, `trino_meta()`) is unchanged and verified identical to Trino.

### description.yml

- `ref` → `b181c61…`, `version` → `0.3.0`.
- `docs.hello_world` / `description` / `extended_description` corrected: the
  previous text advertised `'STRASSE'` and `'i̇stanbul'`, which is what the old
  binary produced but not what Trino does.

### CI

Upstream MainDistributionPipeline on `b181c61` is green on the full matrix
against DuckDB v1.5.5 (Linux amd64/arm64, macOS amd64/arm64, Windows MSVC +
MinGW, Wasm mvp/eh/threads, Format + Tidy):
https://github.com/brikk/duckdb-trino-parity-extension/actions/runs/33609826315

### DIFF
--- a/extensions/trino_parity/description.yml
+++ b/extensions/trino_parity/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: trino_parity
-  description: Native Trino-compatible scalar functions (trino_*) for the cases where DuckDB's built-ins diverge from Trino on Unicode and byte-level inputs — ICU full case folding, code-point reverse, Java-whitespace trim, NFC normalize, and raw-bytes xxhash64/sha512/hmac_sha256
-  version: 0.2.0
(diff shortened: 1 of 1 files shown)


lghavami:
Thanks!

PR #2596, opened 2026-09-02
Add duckdb-dynamodb extension

lukaswelsch:
Adds an extension that enables DuckDB to query DynamoDB tables via a dynamodb_scan function.


lghavami:
Thanks!

PR #2598, opened 2026-09-03
duck_block_utils: v2.0.0

teaguesterling:
MAJOR release. Every `db_*` function was renamed to `duck_block_*` (builders) or `duck_blocks_*` (operations on a list). 54 names present in 1.6.1 are gone with no alias path, so existing queries must be updated. The mapping is mechanical — `db_heading` → `duck_block_heading`, `db_assemble` → `duck_blocks_assemble`, and `db_page` → `page_break`.

**The listing docs needed rewriting, not just the pin.** `hello_world` called `db_assemble`, `db_heading`, `db_paragraph`, `db_text`, `db_bold`, `db_code` and `db_blocks_toc` — none of which exist in 2.0.0 — and the extended description named several dozen more. Both are rewritten, the runnable statements were executed against the built extension rather than transcribed, and every function the listing names was checked to exist: 44 of 44.

Also corrected in the type description, which had drifted from the spec: `kind` is `'block'`, `'inline'` or `'value'` (the third was missing), `level` is depth in a depth-first ordering rather than a heading rank, `content` is populated iff the element has a single text child, and there are eight encodings rather than five.

**andium is deliberately left at its v1.2.1 pin.** The tree this release builds from targets DuckDB v1.5.5 and carries no v1.4.x support, so advancing that pin would point a v1.4.5 build at a tree that cannot build for it.

Highlights beyond the rename — a page axis (`duck_blocks_page_rows`, `duck_blocks_get_pages`) so page boundaries are addressable without synthesising headings; `duck_blocks_diff` and `duck_blocks_quality`; and fixes for a native table exporting a shape pandoc refuses, inline images losing their alt text, and a flattener that emptied formatted table cells.

Release notes: https://github.com/teaguesterling/duckdb_duck_block_utils/releases/tag/v2.0.0

### DIFF
--- a/extensions/duck_block_utils/description.yml
+++ b/extensions/duck_block_utils/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: duck_block_utils
   description: Build, transform, validate, and extract content from structured documents using the duck_block type
-  version: 1.6.1
+  version: 2.0.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


lghavami:
Nice, thanks!

PR #2604, opened 2026-09-03
Add panduck: native document readers into the duck_block vocabulary

teaguesterling:
`panduck` reads rich document formats **natively** into the shared `duck_block`
vocabulary — no Pandoc binary required — so a Word document, an EPUB and a LaTeX
paper all land in the same seven-column shape and can be queried together.

```sql
LOAD panduck;

SELECT * FROM read_docx_blocks('report.docx');
SELECT * FROM panduck_read_blocks('notes.org');   -- reader chosen by path
SELECT * FROM doc_toc('paper.tex');

-- many documents at once, with provenance
SELECT filename, element_type, content
FROM read_panduck_doc('papers/*.docx', filename := true);

-- and back out as Pandoc JSON a real pandoc accepts
SELECT panduck_write_pandoc_ast('doc.json',
    (SELECT list(b) FROM read_odt_blocks('doc.odt') b));
```

**Eleven native readers** — DOCX, ODT, EPUB, RTF, LaTeX, Org, RST, ipynb, MediaWiki,
Textile and PDF — plus a Pandoc AST reader that reaches every format pandoc reads,
for people who already have it.

### Why not link Pandoc

Pandoc is a Haskell program. The only C bindings that ever existed went
unmaintained in 2017, and a Cabal foreign-library request has been open upstream
since 2020. Linking the GHC runtime into a `dlopen`'d DuckDB extension would be a
bad neighbour even if it existed. panduck is compatible with Pandoc's **data
model** instead: every reader is hand-written, and a checked mapping table pins
the correspondence to Pandoc's AST constructors against a real pandoc in CI.

### Ecosystem

Uses the same `duck_block` type as
[`duck_block_utils`](https://github.com/teaguesterling/duckdb_duck_block_utils),
[`markdown`](https://github.com/teaguesterling/duckdb_markdown) and
[`webbed`](https://github.com/teaguesterling/duckdb_webbed), and routes `.md`,
`.html` and `.pdf` to those rather than reimplementing them.

> **Dependency note for reviewers.** `doc_toc` requires **duck_block_utils ≥ 3.0.0**
> (spec 6.5). 3.0.x reshaped `duck_blocks_toc` to return `duck_block[]` and moved the
> projection `doc_toc` reads to `duck_blocks_toc_structs`, so against an older build
> `doc_toc` raises rather than returning wrong rows. duck_block_utils v3.0.1 is
> [#2649](https://github.com/duckdb/community-extensions/pull/2649), merged. Every
> other panduck function is independent of it.

### Testing

**2267 assertions**, 38 fixtures, eight checks, and fourteen CI jobs green across
Linux, macOS, Windows and WASM — plus a forward-compat canary that builds against
DuckDB `main` (the v2.0 line), currently green. That canary is not decoration: it
caught a v2.0 source break in this extension (`ScalarFunction::null_handling` moved
behind a setter) that the v1.5.5 pin compiles clean, which `build_next.yml` would
otherwise have found here.

CI also **inspects the built `.wasm`** for unresolved dependency symbols — CI builds
wasm but never instantiates it, so a missing `LINKED_LIBS` entry otherwise ships
green and throws on first call in the browser. Three sibling extensions shipped that
bug; this one caught it before release.

The readers were audited over five rounds before submission, each round asking a
question the previous could not answer — does a construct come back at all, do any
words go missing, do all readers agree on one logical document. That found 18
defects, including silent text loss in DOCX. All are fixed and pinned by tests.

**Known gaps, stated rather than left for discovery.** `doc_toc` has no CI coverage
on this tag: its test file is gated until the community registry actually *serves*
duck_block_utils 3.0.x, and it is verified instead against a local 3.0.0 build.
Tracked as [#22](https://github.com/teaguesterling/duckdb_panduck/issues/22). The
reader-policy settings do not cover panduck's own readers called directly; that is
[#16](https://github.com/teaguesterling/duckdb_panduck/issues/16), fixed on `main`
and shipping next. Both are documented in the extension's own docs.

### Notes on the descriptor

- Ref is [`v0.2.0`](https://github.com/teaguesterling/duckdb_panduck/releases/tag/v0.2.0)
  (`a57402b`) — an annotated tag with a published release and full changelog.
- No `andium` ref — panduck targets DuckDB v1.5.5 and has never been
  build-verified against the v1.4.5 track, so naming a commit there would claim
  something untested.
- No `docs_url` — the documentation site is not published yet, and a dead link is
  worse than none.
- `vcpkg_commit` is the `extension-ci-tools` v1.5.5 default, which this extension's
  own pipeline builds green against on every platform.

### DIFF
--- a/extensions/panduck/description.yml
+++ b/extensions/panduck/description.yml
@@ -13,7 +13,7 @@ extension:
   vcpkg_commit: 84bab45d415d22042bd0b9081aea57f362da3f35
 repo:
   github: teaguesterling/duckdb_panduck
-  ref: 67618f67290ecd35bc9ec0663e942331f22d4a5d
+  ref: dedc85c5fe0bc59aadc3019659574a7e283731e4
 docs:
   hello_world: |
     LOAD panduck;
--- a/extensions/panduck/description.yml
+++ b/extensions/panduck/description.yml
@@ -1,0 +1,125 @@
+extension:
+  name: panduck
+  description: Read documents natively into the duck_block vocabulary -- DOCX, ODT, EPUB, RTF, LaTeX, Org, RST, ipynb, MediaWiki and Textile -- and write them back as Pandoc JSON, without linking Pandoc
+  version: 0.1.1
+  language: C++
+  build: cmake
+  license: MIT
+  requires_toolchains: vcpkg
+  maintainers:
+    - teaguesterling
+  # The default pin from extension-ci-tools v1.5.5, which is what panduck's own
+  # distribution pipeline builds green against on every platform including wasm.
+  vcpkg_commit: 84bab45d415d22042bd0b9081aea57f362da3f35
+repo:
+  github: teaguesterling/duckdb_panduck
+  ref: a2fdc54d8be3132f7eb31c3bff6679418ff7cdfa
+docs:
+  hello_world: |
+    LOAD panduck;
+
+    -- Read a document straight into flattened duck_block rows
+    SELECT * FROM read_docx_blocks('report.docx');
+    SELECT * FROM read_epub_blocks('book.epub');
+    SELECT * FROM read_latex_blocks('paper.tex');
+
+    -- Or let the path pick the reader
+    SELECT * FROM panduck_read_blocks('notes.org');
+    SELECT panduck_format_for('paper.tex');       -- 'latex'
+    SELECT panduck_can_read('archive.docx');      -- true
+
+    -- Formats with no file: parse a string directly
+    SELECT * FROM read_mediawiki_blocks_string('== Heading ==');
+
+    -- Outline any supported document
+    SELECT * FROM doc_toc('report.docx');
+
+    -- Write back out as Pandoc JSON that a real pandoc accepts
+    SELECT panduck_write_pandoc_ast('doc.json',
+        (SELECT list(b) FROM read_odt_blocks('doc.odt') b)
+    );
+
+    -- Or get the AST as a struct without touching the filesystem
+    SELECT panduck_blocks_to_pandoc_ast(
+        (SELECT list(b) FROM read_rtf_blocks('memo.rtf') b)
+    );
+
+  extended_description: |
+    Panduck reads document formats natively into the `duck_block` vocabulary shared
+    with `duck_block_utils`, `markdown` and `webbed` -- so a DOCX, an EPUB and a
+    LaTeX paper all land in the same seven-column shape and can be queried together.
+
+    **It does not link Pandoc.** Pandoc is a Haskell program; the only C bindings
+    that ever existed went unmaintained in 2017, and a Cabal foreign-library request
+    has been open upstream since 2020. Linking the GHC runtime into a dlopen'd DuckDB
+    extension would be a bad neighbour even if it existed. Panduck is instead
+    compatible with Pandoc's **data model**: every reader is hand-written, and a
+    checked mapping table pins the correspondence to Pandoc's AST constructors.
+
+    ## Readers
+
+    | Function | Format |
+    |----------|--------|
+    | `read_docx_blocks(path)` | Word OOXML |
+    | `read_odt_blocks(path)` | OpenDocument Text |
+    | `read_epub_blocks(path)` | EPUB 2/3 |
+    | `read_rtf_blocks(path)` | Rich Text Format |
+    | `read_latex_blocks(path)` | LaTeX |
+    | `read_org_blocks(path)` | Org mode |
+    | `read_rst_blocks(path)` | reStructuredText |
+    | `read_ipynb_blocks(path)` | Jupyter notebooks |
+    | `read_mediawiki_blocks(path)` | MediaWiki wikitext |
+    | `read_textile_blocks(path)` | Textile |
+    | `read_pandoc_blocks(path)` | Pandoc JSON AST (reaches all 43 formats pandoc reads) |
+
+    Every text format also has a `_string` variant that parses content directly.
+    DOCX, ODT, EPUB and RTF read lists, blockquotes, tables, images and footnotes;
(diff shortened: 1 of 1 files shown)


teaguesterling:
Updated to `0.1.2` / `ref: e4d3115`, which is tagged `v0.1.2` upstream. panduck's own CI is green on that commit — 14 jobs across Linux, macOS, Windows and WASM.

**On the failing `test_against_latest` check** — flagging it deliberately rather than leaving a reviewer to wonder.

All `build` jobs pass against the pinned `v1.5.5` this extension targets. The two failures are from the job that builds against the `v2.0-cyanoptera` pre-release. Four API changes affect panduck there; two are fixed in `0.1.2`, and both were fixed by *removing* a version assumption rather than branching:

- the bind callback's `names` parameter (`vector<string>&` → `vector<Identifier>&`) — the type is now derived from DuckDB's own `table_function_bind_t`, so it follows whatever DuckDB declares
- `ScalarFunction::null_handling` → `SetNullHandling()` — selected by overload resolution, no version named

Two are **not** fixed, deliberately:

- `UnaryExecutor::ExecuteWithNulls` is gone, and v2.0 changed the null *protocol* to an empty `optional<T>` return. Writing the loop directly against the vector API avoids the executor, but v2.0 also renamed the mutable accessors (`GetData`→`GetDataMutable`, `Validity`→`ValidityMutable`), so the output side still needs a per-accessor shim.
- `DefaultMacro` collapsed from five fields to three, folding the parameter list into the definition string. The struct *shape* differs, so the eight scalar macros can't be written once. (`DefaultTableMacro` is unchanged.)

Since v2.0 needs both, doing only the tractable half would add version shims while gaining nothing today. I'd rather port these once v2.0 ships and its API has settled than carry two code paths against a moving pre-release — but happy to do it now if you'd prefer the check green before merge.

teaguesterling:
Updated to `0.1.4` / `ref: c00bf4d`, tagged `v0.1.4` upstream. **All checks green** — the pinned v1.5.5 build and the `v2.0-cyanoptera` forward-compatibility job, across Linux, macOS, Windows and WASM.

Following up on my earlier note: v2.0 support is now complete rather than partial. Four API changes affected this extension, and all are handled without naming a version anywhere:

| Change | Approach |
|---|---|
| bind `names`: `vector<string>&` → `vector<Identifier>&` | type derived from `table_function_bind_t`, so it follows whatever DuckDB declares |
| `ScalarFunction::null_handling` → `SetNullHandling()` | selected by overload resolution |
| `UnaryExecutor::ExecuteWithNulls` removed | replaced with `Vector::SetValue`, identical in both versions |
| `DefaultMacro` 5 fields → 3 | neutral macro table; only the conversion is version-aware |

### One observation you may want to know about

While getting there I hit something that looks like a v2.0 issue rather than an extension one, so flagging it in case it's useful — I have not filed it anywhere.

A `WHERE` predicate over a table macro appears not to be applied under v2.0. The same filter returned nearly every row when the query aggregated with `string_agg`, and zero rows when it aggregated with `count(*)` — same data, same predicate:

```
v1.5.5, NO filter:    inline/italic, block/paragraph, inline/text,
                      inline/strikethrough, inline/bold, block/heading
v2.0,   WITH filter:  inline/bold, block/paragraph, inline/italic,
                      block/heading, inline/text
```

The v2.0 "filtered" result is the v1.5.5 *unfiltered* set less one element, and every row it returned carried `kind='inline'`/`'block'` and `encoding='text'` — satisfying neither disjunct of the predicate that selected it. The projection printed correct values throughout, so the rows themselves were fine.

I did **not** change any reader for this. The affected assertion now evaluates its property in the projection via `bool_and` instead of filtering, which doesn't depend on filter planning and is stricter (an empty result fails rather than passing vacuously). If that assertion looks oddly written, that's why.

Happy to dig further or move it somewhere more appropriate if it's worth a look.

teaguesterling:
Bumped to **0.1.5** (`ref` → `2d9c3b2`, tagged `v0.1.5`). Flagging it since the change is not cosmetic: **0.1.4 would have shipped broken.**

`duck_block_utils` published a rename of its whole surface (`db_*` → `duck_block_*` / `duck_blocks_*`). panduck's `doc_toc` and `doc_render` name those functions *by name* from generated SQL at runtime, so they follow whatever build the user has `INSTALL`ed rather than anything pinned at build time. On 0.1.4, anyone with a current `duck_block_utils` gets:

```
D SELECT count(*) FROM doc_toc('README.md');
Catalog Error: Scalar Function with name db_blocks_toc does not exist!
Did you mean "duck_blocks_toc"?
```

0.1.5 renames the two runtime names panduck holds, and flips the test that detects this in the same commit — the detector's expected value has no meaning apart from the spelling the call sites use.

Also carried since 0.1.4, all green on the pinned v1.5.5 too:

- The three write functions call `SetFallible()`. v2.0 replaces the error of a throwing scalar that hasn't declared it can throw — unconditionally, not under an assertion — which turned panduck's clean nesting-depth error into an `INTERNAL Error`.
- `duckdb_compat.hpp` probes for the member it needs rather than for a header that moved near it in the same refactor. Those two demonstrably separate: `identifier.hpp` was backported to `v1.5-variegata` without the signature change that made it interesting.
- A forward-compat canary builds against DuckDB `main` on pushes to the repo's own `main` — advisory, not a merge gate — so source breaks against the v2.0 line surface there instead of on a release PR here.

Verified on `2d9c3b2`: 2034 assertions across 18 test files, all eight repo checks green, CI green on Linux, macOS, Windows and WASM. No reader behaviour changed.

teaguesterling:
@sebastiaan-dev sorry for the explicit tag but do you mind also reviewing and merging this one? There were some features I moved out of one extension into this new one.

sebastiaan-dev:
No worries, merged!

PR #2610, opened 2026-09-04
ai: bump to v0.4.25

leonardovida:
Update the `ai` extension to v0.4.25 so community installs include the current Databricks completion and provider maintenance fixes.

### DIFF
--- a/extensions/ai/description.yml
+++ b/extensions/ai/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: ai
   description: AI functions for SQL — completions, classification, extraction, embeddings, and read-only SQL generation across local and hosted model providers
-  version: 0.4.14
+  version: 0.4.24
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


leonardovida:
The supported DuckDB 1.5.5 community matrix is green across Linux AMD64/ARM64, macOS AMD64/ARM64, Windows MSVC, and MinGW. The `v2.0-cyanoptera` advisory lane fails during compilation because DuckDB 2.0 is migrating several C++ extension APIs, beginning with table-function bind names changing from `vector<string>` to `vector<Identifier>` and continuing through scalar/aggregate bind and vector-access contracts. I reproduced the full failure locally and avoided adding a partial compatibility shim. The pinned v0.4.25 release and its 13-job DuckDB 1.5.5 distribution are green: https://github.com/leonardovida/duckdb-ai/actions/runs/33856718504

sebastiaan-dev:
Thanks!

PR #2611, opened 2026-09-04
oracle_scanner: bump to 0.2.1

krokozyab:
Bumps oracle_scanner from 0.2.0 to 0.2.1.

Fixes a hard failure on wide tables: an Oracle DESCRIBE_INFO that does not fit
one packet made 0.2.0 stop at column 120 with "TTC message is truncated"

Fix contributed by @pindamonhangaba


krokozyab:
The DuckDB 1.5.5 matrix this release targets is green across Linux
amd64/arm64, macOS arm64/x86_64 and Windows MSVC:
https://github.com/krokozyab/quack-oracle/actions/runs/33861601152

The `test_against_latest` lane fails at compile time because DuckDB main is
migrating several C++ extension APIs. In this extension it lands in two files:
table filters move from `TableFilter`/`TableFilterType::OPTIONAL_FILTER` to
`ExpressionFilter` with a `BoundFunctionExpression`, catalog identifiers change
from `std::string` to `Identifier`, `CreateTableInfo::schema/table` become
`SetSchema()`, and `Timestamp::IsFinite` is no longer a static.

I reproduced the failure locally and deliberately did not add a partial
compatibility shim: filter translation is where this extension guarantees that
a pushed-down predicate means the same thing to Oracle as it does to DuckDB, and
maintaining two implementations of it against an API that is still moving would
put that guarantee behind an #if. The port lives on a dedicated branch instead,
with a weekly advisory workflow that compiles it against DuckDB main and reports
the size of the gap without failing the repository.

sebastiaan-dev:
Hi @krokozyab, this was an issue on our end where CI was misconfigured. I'll rerun your PR.

sebastiaan-dev:
Merged!

PR #2614, opened 2026-09-05
trino_parity: bump to 0.4.0 (reject empty HMAC keys like Trino)

apatrida:
Bumps `trino_parity` to `e93c6b78f460db4f1fc93b78f4b771a8665e3ec0`
(version 0.4.0).

### Fix

`trino_hmac_sha256(data, key)` now rejects a zero-byte key with
`InvalidInputException("Empty key")`, matching Trino's `SecretKeySpec` API
contract. RFC 2104 itself permits an empty key, so the previous implementation
returned a digest where Trino fails. This became user-reachable once duckbridge
added BLOB→VARBINARY table-column mapping.

Empty **data** remains valid. Sqllogic also pins binary-NUL keys/messages,
>64-byte keys (hash-the-key path), and NULL propagation on either argument.
No catalog/function signatures changed.

### DuckDB latest compatibility

The first `test_against_latest` run exposed dead pre-0.2 scalar-macro machinery:
the extension had no scalar macros after its native-function shrink, but still
compiled a sentinel-only `DefaultMacro` array and no-op registration loop.
DuckDB latest changed that unused API. Commit `e93c6b7` deletes the empty array,
loop and dependency rather than adding a version shim. `trino_meta()` remains on
its separate `DefaultTableMacro` path; all ten `trino_*` functions remain native
scalars.

### CI

The HMAC change's upstream MainDistributionPipeline is green across Linux
amd64/arm64, macOS amd64/arm64, Windows MSVC + MinGW, Wasm mvp/eh/threads,
Format and Tidy:
https://github.com/brikk/duckdb-trino-parity-extension/actions/runs/33933840257

The final ref reruns that matrix. Local pinned-DuckDB v1.5.5 format/build/tests:
82/82 assertions.

### DIFF
--- a/extensions/trino_parity/description.yml
+++ b/extensions/trino_parity/description.yml
@@ -1,7 +1,7 @@
 extension:
   name: trino_parity
   description: Native Trino-compatible scalar functions (trino_*) for the cases where DuckDB's built-ins diverge from Trino on Unicode and byte-level inputs — Trino's simple per-code-point case mapping, code-point reverse, Java-whitespace trim, NFC normalize, and raw-bytes xxhash64/sha512/hmac_sha256
-  version: 0.3.0
+  version: 0.4.0
   language: C++
   build: cmake
   license: MIT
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Hi, I merged your v0.5.0 PR, so I'll close this one in favor of the newer version.

PR #2647, opened 2026-09-06
bump marc21 to v1.0.1

mgbilby:
Bumps marc21 from v1.0.0 to v1.0.1.

    ref points at v1.0.1 in mgbilby/duckdb-marc21.
    Distribution pipeline green on Linux, macOS, WASM and Windows at that commit.


sebastiaan-dev:
Thanks!

PR #2654, opened 2026-09-06
Update DuckFlight to 0.1.5

nicosuave:
Update DuckFlight to 0.1.5 with PostgreSQL client/catalog compatibility and Flight SQL schema fixes. The public shim embeds checksum-pinned core-v0.1.4 binaries for the existing four supported platforms.


sebastiaan-dev:
Thanks!

PR #2661, opened 2026-09-06
duck_tails: v1.6.0

teaguesterling:
DuckDB 2.0.0 compatibility.

- `GitFileSystem` implements the positional `Read(handle, buffer, nr_bytes, location)`. It previously overrode only the sequential form; v1.5's readers call the sequential one and v2.0's call the positional one, so every read through `git://` failed on v2.0. 13 of 59 test files failed before, 59 pass after.
- Fixture setup gated on `SKIP_TESTS`, since linux runs `make test_<type>` twice.
- Retired a temporary `post_build_command` that hooked a target extension-ci-tools no longer defines.
- CI: the duckdb-main canary was failing at repository discovery and masking the above; fixed by declaring git `safe.directory` at system level, which libgit2 resolves without `$HOME`.


lghavami:
Thanks!

PR #2666, opened 2026-09-07
Updates cozip to 2.0.0

csaybar:
Added

- read_taco(path) reads a hierarchical dataset stored in a cozip archive, returning one row per sample with one column per file, each column holding a GDAL /vsisubfile/ pointer. Pass pivoted := false for one row per file.
- The same function handles the three container shapes, a .zip archive, a plain directory, and a directory with consolidated metadata.
- idx, files and level narrow the scan before the pivot, and taco_contract(), taco_structure(), taco_levels() and taco_collection() expose the dataset layout without loading the json extension.
- Test fixtures for flat, hierarchical, variable-leaf and null-structure archives, plus the directory containers.

Changed

- read_cozip is now read_flat. The old name stays as a deprecated alias and will be removed in 3.0.
- read_flat reports which profile it found when handed the wrong kind of archive.
- The byte-0 index parser now returns every entry from a single read and rejects entries whose byte range falls outside the archive.
- COLLECTION.json is parsed with the yyjson DuckDB already vendors, instead of a hand-written scanner.

Removed

- WebAssembly builds.

### DIFF
--- a/extensions/cozip/description.yml
+++ b/extensions/cozip/description.yml
@@ -1,31 +1,40 @@
 extension:
   name: cozip
   description: Cloud-Optimized ZIP reader for DuckDB
-  version: 1.2.0
+  version: 2.0.0
   language: C++
   build: cmake
   license: MIT
+  excluded_platforms: "wasm_mvp;wasm_eh;wasm_threads"
   maintainers:
     - csaybar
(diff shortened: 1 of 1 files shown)


lghavami:
Thanks!

PR #2671, opened 2026-09-07
oracle_scanner: bump to 0.2.2

krokozyab:
Databases whose character set is not AL32UTF8 were refused at connect. The ref is the commit tagged v0.2.2, whose matrix is green on Linux amd64/arm64, macOS arm64/x86_64 and Windows.


sebastiaan-dev:
Thanks!

PR #2672, opened 2026-09-08
Update cozip to 2.0.1

csaybar:
Hi, this is a small bug fix for `read_taco`.

```sql
SELECT path
FROM read_taco(
  'dataset.zip',
  pivoted := false,
  files := ['mask.tif']
);
```

Before, this returned every file. Now it returns only `mask.tif`. Unknown file names now raise an error instead of being ignored.

Thanks!


sebastiaan-dev:
Thanks!

PR #2680, opened 2026-09-09
ossie: bump to 0.1.1

venkata-chikkam:
Bumps `ossie` to 0.1.1 and points `ref` at the [v0.1.1](https://github.com/iqea-ai/duckdb-ossie/releases/tag/v0.1.1) tag (`5180c67`).

A patch follows 0.1.0 quickly because **0.1.0's `extended_description` contained a security claim that was false**, and I would rather correct it than leave it rendered on the extension page.

### What was wrong

0.1.0 said an agent connected over MCP *"can reach nothing but the model's own vocabulary."* `duckdb_mcp` also publishes generic `query`, `export`, `list_tables`, `describe` and `database_info` tools; calling `query` returned 7,212 rows from a table the semantic model never declares.

`examples/server.sql` now disables all five when it starts the server, so `tools/list` returns only `semantic_query` and calling anything else is refused with *Tool not found*. That makes the original claim true, and a CI job asserts it on every push — including a control server started *without* the flags, so the assertion cannot pass merely because the built-ins stopped being published.

Reported upstream as teaguesterling/duckdb_mcp#75; the per-tool flags existed but were missing from that project's tool table, which the maintainer has since fixed. Worth noting for anyone else publishing a curated tool: disabling `query` alone is not sufficient, because `export` takes an arbitrary SQL argument and reaches the same data.

### Also in this release

- `examples/server.sql` never loaded the extension it demonstrates — it loaded `duckdb_mcp` but not `ossie`, and called `dsdgen` without `tpcds`, so it only worked from a build tree. Following the README after a registry install hit `Catalog Error: Table Function with name ossie_load does not exist!`
- Dropped `-unsigned` from the documented invocation and the Claude Desktop config snippet; community extensions are signed, so it only disabled signature verification for the whole session
- CI now loads the built `.duckdb_extension` on `linux_amd64` and `osx_arm64` and runs a 40-assertion end-to-end check against it, rather than only testing inside the build tree

### Two documentation claims also corrected

While verifying everything in this descriptor before submitting, two statements turned out to be wrong:

- `core-spec/osi-schema.json` → `core-spec/ossie-schema.json`. Apache renamed it; the old path 404s
- *"checked against hand-written TPC-DS SQL at sf=1 across every metric and every dimension"* — the differential tests run `dsdgen(sf = 0.01)` and cover 2 of 5 metrics. Now stated accurately

### Verification

- Built from `5180c67` through this repo's own pipeline on a fork: 11/14 jobs success, 0 failures (`deploy`/`archive`/`doc_test` skipped off `main`) — https://github.com/iqea-ai/community-extensions/actions/runs/34392215844
- `scripts/build.py` run against this descriptor resolves to `COMMUNITY_EXTENSION_REF=5180c67299ca14fde9d46806c7d35f9f53c2f1a9`
- `hello_world` executed by hand: returns `US|350.00`, `DE|75.00`
- Upstream `iqea-ai/duckdb-ossie`: 16/16 CI green at the tagged commit

### DIFF
--- a/extensions/ossie/description.yml
+++ b/extensions/ossie/description.yml
@@ -3,7 +3,7 @@ extension:
   description: The Apache Ossie (incubating) reference implementation for DuckDB. Reads Ossie
     semantic model files in YAML or JSON and answers semantic queries against the tables in DuckDB,
     from a single binary, with no infrastructure. Also exposes the semantic layer via MCP.
-  version: 0.1.0
+  version: 0.1.1
   language: C++
   build: cmake
   license: MIT
@@ -13,7 +13,7 @@ extension:
 
 repo:
   github: iqea-ai/duckdb-ossie
-  ref: ef995aa581455840c7adc4e24d30220b047962a3
+  ref: 5180c67299ca14fde9d46806c7d35f9f53c2f1a9
 
 docs:
   hello_world: |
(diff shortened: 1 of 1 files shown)


sebastiaan-dev:
Thanks!

PR #2687, opened 2026-09-13
duck_block_utils 3.2.0

teaguesterling:
Bumps duck_block_utils to v3.2.0 (tag on 079123d). Spec duck_blocks 1.3, additive: the body predicate (`duck_block_is_body`), and `duck_blocks_to_text` no longer renders a frontmatter/tailmatter metadata blob as prose. Release notes: https://github.com/teaguesterling/duckdb_duck_block_utils/releases/tag/v3.2.0

`ref_next` set to the same commit. `andium` pin unchanged, as before.

PR #2688, opened 2026-09-13
panduck 0.5.0

teaguesterling:
Bumps `panduck` to **0.5.0**, tagged and released at
[`5a0331b`](https://github.com/teaguesterling/duckdb_panduck/releases/tag/v0.5.0).

`ref` and `ref_next` are the same commit deliberately — without `ref_next`,
`scripts/build.py` prints "Skipping prerelease validation" and the PR goes green having
never built against DuckDB-next.

**What's in it**

- `expand_embedded := true` on `read_panduck_doc`, `doc_toc`, `doc_section`,
  `doc_search_sections` and `doc_container`. A `.ipynb` markdown cell is held as one `raw`
  block, so a notebook had no headings and nothing for `doc_toc` to work with; the
  expansion happens in the SQL dispatch layer so the C++ reader keeps its isolation. Opt-in,
  and an unchanged call reads identically without the `markdown` extension present.
- Bare `list_item`, `caption` and `inline` blocks now export as `BulletList`/`Figure`/`Plain`
  instead of producing `"blocks": []` or being dropped.
- `panduck_blocks_to_pandoc_json(blocks)`; the old `panduck_pandoc_ast_json` still works.
- `.rst` and `.textile` table cells resolve their inline markup.
- Vocabulary re-vendored at duck_block_utils v3.1.0 (`6c1c2e5`), `SPEC_VERSION 1.2`.

**Descriptor corrections in the same commit.** The Document helpers table listed only
`doc_toc` and `doc_render` while the `hello_world` directly above it already demonstrated
`doc_section` and `doc_search_sections`; the assertion count (2384) and the CI check count
("eight") were both stale. Those now state what was measured.

Sibling PR: #2687 (`duck_block_utils` 3.2.0) — same fleet, same batch.

https://claude.ai/code/session_019nkprWWWSkwuWNmAEoVUs2


lghavami:
Thanks!

PR #2689, opened 2026-09-13
Introduce px-extension

toppyy:
Adds the [px-extension](https://github.com/toppyy/duckdb-px) for querying [Px-files](https://www.scb.se/en/services/statistical-programs-for-px-files/px-file-format/) that various statistical agencies use for publishing tabular data.


lghavami:
Thanks!

PR #2690, opened 2026-09-13
markdown 1.9.1

teaguesterling:
`markdown` 1.8.0 → **1.9.1**, at [`2ba1321c`](https://github.com/teaguesterling/duckdb_markdown/commit/2ba1321c5c2b97f55d667a9aab4c0e55e1bfbf22) ([v1.9.1 notes](https://github.com/teaguesterling/duckdb_markdown/releases/tag/v1.9.1), [v1.9.0 notes](https://github.com/teaguesterling/duckdb_markdown/releases/tag/v1.9.0)).

`ref_next` is set, and it equals `ref`.

## This PR first pointed at 1.9.0, which failed here

The earlier revision targeted v1.9.0 (`bd77b934`). Its `test_against_latest` leg failed on **both** Linux architectures:

```
FAILED: src/markdown_reader_files.cpp.o
error: use of deleted function
  'duckdb::ConstantExpression::ConstantExpression(const duckdb::Value &)'
```

DuckDB 2.0 deletes that constructor and replaces it with `ConstantExpression::FromValue`. v1.9.1 calls `FromValue` wherever it exists, so the code compiles on both lines. v1.9.1 has no behaviour changes relative to v1.9.0.

**That failure was the first time markdown had ever been built against the next line.** The descriptor had no `ref_next` before this PR, so every earlier release printed "Skipping prerelease validation" and passed without the next-line build ever running. Rather than drop `ref_next` to get a green check, this PR now targets a release that actually builds.

## Changes visible to consumers (from 1.9.0)

- **Frontmatter no longer renders as prose** in `md_to_text` or `md_to_html`. A document whose entire content is a frontmatter block now returns empty.
- **`element_order` is dense from 0.** It used to start at 1. `webbed` and `panduck`'s native readers already started at 0, so this removes a cross-producer inconsistency rather than introducing one. `panduck`'s `.md` path delegates to markdown and changes with it.
- **Tight and loose lists survive both reading and writing.** Every document containing a list changes block shape: in a tight list, the text now sits on `list_item` rather than in a `paragraph` child.
- **Blockquotes keep their paragraph.** A blockquote's own `content` is now NULL, and its text sits one level deeper.

## Verification

- **Stable (v1.5.5):** every platform built on `2ba1321c`.
- **Next line (DuckDB 2.0), this PR's `test_against_latest`:** Linux amd64 and arm64, macOS amd64 and arm64, Windows amd64 and mingw, and all three WebAssembly variants passed. On v1.9.0, both Linux legs failed here with the deleted constructor.
- **Next line, the repository's own CI:** pull requests build a reduced set (Linux amd64, Windows amd64 and `wasm_eh`), and pushes to `main` build the full set. The push that merged this release scheduled all nine platforms, the same matrix the registry builds.
- Full suite: 1948 assertions in 54 cases. `check-conformance` and `check-vocabulary` are clean.

## Notes for review

- `andium` is deliberately unchanged. That track is pinned to a v1.4.5 tree, and every change since then ships on the v1.5.x track via `ref`.
- The repository's DuckDB submodule is now pinned at exactly `v1.5.5` (`d8cdaa33fd`), the version this registry builds with, and a CI job fails if the two ever diverge.


lghavami:
Thanks!

PR #2691, opened 2026-09-13
Update DuckFlight to 0.1.6

nicosuave:
Update DuckFlight to 0.1.6, using the exact public extension commit with the core-v0.1.5 runtime. The runtime supplies catalog comments for PostgreSQL compatibility macros and internal helpers, and the extension documentation describes the five native server-control functions.


lghavami:
Thanks!

PR #2700, opened 2026-09-14
Update sshfs to DuckDB v1.5.5

onnimonni:
Bumps sshfs to DuckDB v1.5.5.

- `ref` now points to the tip of the extension's `v1.5` branch: duckdb/extension-ci-tools pinned to v1.5.5, built and tested against DuckDB v1.5.5 (distribution pipeline + SFTP integration tests green).
- Also picks up the SSH KEX handshake fix and a command-injection fix that were previously only on `main` (v1.4 line).
- `andium` ref for the v1.4.5 LTS line is unchanged.

https://claude.ai/code/session_019K9Begbj9inUdA1dCQPerB


lghavami:
Thanks!

PR #2702, opened 2026-09-14
markdown 1.9.2

teaguesterling:
`markdown` 1.9.1 → **1.9.2**, at [`ee544a8`](https://github.com/teaguesterling/duckdb_markdown/commit/ee544a8a77ecae13775adb729eeb987dad9ece67) ([v1.9.2 notes](https://github.com/teaguesterling/duckdb_markdown/releases/tag/v1.9.2)). `ref_next` is set, and it equals `ref`.

## What this release fixes

On DuckDB 2.0, `md_extract_table_rows` and `md_extract_tables_json` raised an INTERNAL error when a table cell nested past the extension's inline depth cap of 1000. They now raise the intended error:

```
Invalid Input Error: Markdown inline nesting exceeds maximum supported depth (1000)
```

Both functions reach an `InvalidInputException` and were not marked `SetFallible()`. DuckDB 2.0 rewrites an execution error thrown by an unmarked scalar function into an `InternalException`. The fix marks both functions.

The v1.5 line has no behaviour change.

## Verification

- **A new test pins the error class.** Depth 1000 returns the table, and depth 1001 must fail with `Invalid Input Error: `.
  - **Before the fix:** pushed alone, the test failed on the DuckDB 2.0 build with the INTERNAL form (54 passed, 1 failed).
  - **With the fix:** 55 passed.
- **Stable (v1.5.5):** 1952 assertions in 55 test cases. The only difference from v1.9.1 is the new test.
- **Release commit, repository CI:** DuckDB 2.0 (`v2.0-cyanoptera`) passed on all nine platforms, the same matrix this registry builds. An advisory build against DuckDB `main` also passed on all nine.

## Notes for review

- `andium` is unchanged, for the same reason as in #2690.
- Since v1.9.1, the extension source differs only by the fix. Every other change on the repository's main since v1.9.1 is CI configuration.


lghavami:
Thanks!

PR #2703, opened 2026-09-15
duck_block_utils 3.3.0

teaguesterling:
Bumps duck_block_utils to v3.3.0 (tag on 95a84e6). Spec duck_blocks 1.4, additive: body is a subtree property, new `duck_blocks_body`, validation rule L6, and `duck_blocks_to_text` level-scopes the metadata blob. Release notes: https://github.com/teaguesterling/duckdb_duck_block_utils/releases/tag/v3.3.0

`ref_next` set to the same commit. `andium` pin unchanged, as before.

This also supersedes 3.2.0 (#2687), whose post-merge build failed on a yum metadata download before compiling and was never deployed; 3.3.0 contains everything in 3.2.0.


lghavami:
Thanks!

PR #2706, opened 2026-09-15
Add lerobot extension for LeRobot datasets

kaka11chen:
## Add the `lerobot` extension

Query and create LeRobot v3 robotics datasets with SQL: read metadata and frame
data, write datasets with `COPY`, decode RGB video, and select observation windows.
The optional TorchCodec Python adapter is a separate package.

The descriptor pins [AstroVela/duckdb-lerobot@901f26a](https://github.com/AstroVela/duckdb-lerobot/commit/901f26a6639c3f3ba763a4df758d83e197694806)
on its default branch, `v1.5-variegata`. It targets DuckDB 1.5.5 and the existing
`v1.5-variegata` CI tools, requests the multimedia toolchain, and excludes Wasm.

## Source distribution

The extension's own source is Apache-2.0. Its default native build statically
links LGPL FFmpeg, libaom, dav1d, and zlib; their licenses also apply. GPL and
nonfree FFmpeg configurations are rejected by the default release packager.

The project's [distribution workflow](https://github.com/AstroVela/duckdb-lerobot/blob/901f26a6639c3f3ba763a4df758d83e197694806/.github/workflows/MainDistributionPipeline.yml)
enables `LEROBOT_PACKAGE_RELEASE=1` and disables dependency binary caches. It
verifies the linked libraries against installed source records and includes a
matching `release/` directory with each platform artifact: dependency sources,
port recipes and patches, licenses, DuckDB and extension source, rebuild tools,
and a manifest containing the unsigned extension's SHA-256.
The Linux rebuild job modifies FFmpeg, relinks the extension, and runs SQL tests.
See [REDISTRIBUTION.md](https://github.com/AstroVela/duckdb-lerobot/blob/901f26a6639c3f3ba763a4df758d83e197694806/REDISTRIBUTION.md).

The community build calls the shared distribution workflow, which currently
uploads only the binary. The separate repository archive removes DuckDB and
does not collect the dependency sources from that build. The project's generated
source bundle therefore needs an explicit path through community publishing.

Proposed integration for review:

1. Enable release packaging in the community build and forward the existing
   shared-workflow `vcpkg_binary_sources: clear` input. Packaging must use the
   actual community build's sources and configuration.
2. Add opt-in artifact support for the accompanying `release/` directory in
   extension-ci-tools, preserving the existing binary's artifact path.
3. In community deployment, check the bundle checksums and its association with
   the unsigned binary, then publish the materials to durable storage before
   publishing the binary. Retain immutable materials per build/platform and
   record the signed/compressed download's checksum and source-bundle location.
4. Expose that source-download location in the extension's community page.

This needs maintainer review of the publishing interface and storage convention.
An expiring Actions artifact or a bundle from a separate build is not the proposed
source-delivery mechanism. The draft retains native video support.


lghavami:
Thanks!

PR #2709, opened 2026-09-15
Bump quack_oauth to v2026.09.15 and exclude Wasm

jrosskopf:
Bumps `quack_oauth` to [`v2026.09.15`](https://github.com/DataZooDE/quack-oauth/releases/tag/v2026.09.15), a security-focused release, and excludes the Wasm platforms.

## Ref bump

Three fixes since `v2026.08.10`:

- **Authorization walk hardening.** Table functions (`read_csv`, `read_parquet`, `glob`, `postgres_query`, `duckdb_secrets`, …) were not gated at all — `TABLE_FUNCTION` refs fell into the "nothing to gate" branch, so `SELECT … FROM entitled, read_csv('http://…')` authorized with only the entitled object walked. They now surface as gated objects under a reserved `fn:<name>` namespace, so default-deny covers them. Separately, expression subqueries were unwalked: `SELECT (SELECT max(x) FROM other) FROM entitled` and `… WHERE EXISTS (…)` reached another schema unchecked. The walk now recurses into `ExpressionClass::SUBQUERY` and visits `where_clause` / `having` / `qualify` / `groups`.
- **JWKS refresh after a cached-key signature failure.** Providers that rotate key material while reusing the same `kid` left the validator holding a stale key. It now refreshes once on failure, rate-limited per `kid`, preserving the last-good key when the IdP is unreachable.
- **Wasm preprocessor guards (#14).** `emcc` defines `__EMSCRIPTEN__`, not bare `EMSCRIPTEN`, so 14 `#ifndef EMSCRIPTEN` guards never fired and every Wasm target failed with `fatal error: 'telemetry.hpp' file not found`. Thanks to @rustyconover for the report from the haybarn community-extensions CI.

## Why also exclude Wasm

The guard fix above makes the Wasm targets *compile*, but the resulting side-module still would not **load**.

`quack_oauth` links OpenSSL (via jwt-cpp) for JWT signature verification. The Wasm loadable is built as an `emcc -sSIDE_MODULE=2` side-module, which links only the libraries named in `duckdb_extension_load(... LINKED_LIBS ...)`, and this extension declares none. The crypto symbols are therefore left unresolved. Because symbol resolution in a side-module is deferred to load time, the `.wasm` builds green in CI and only fails when the browser tries to instantiate it — a failure the build matrix cannot observe.

Rather than publish an artifact that is green in CI and broken for users, we exclude Wasm explicitly, matching the source repository's own distribution matrix. OAuth's interactive and redirect flows are not viable in a Wasm context anyway, so a clean "not available on this platform" is the honest outcome here.

This is not a regression: no working Wasm artifact exists today, since those targets currently fail to compile. The source-side split (`DUCKDB_WASM_SAFE_SOURCES` / `DUCKDB_NATIVE_ONLY_SOURCES` plus the `__EMSCRIPTEN__` guards) is kept so a future `LINKED_LIBS`-based Wasm build can be re-enabled once it can be verified against a real browser load.

## Verification

CI green on the tag in the source repository — 31 jobs across DuckDB v1.5.5 and v1.4.5 LTS, covering `linux_amd64`, `linux_arm64`, `osx_amd64`, `osx_arm64` and `windows_amd64`, including per-platform smoke tests that install the shipped artifact into a stock DuckDB CLI, load it and call a real function.


lghavami:
Thanks!

PR #2711, opened 2026-09-15
subtoken: add new extension to the registry

hughcameron:
Adds `extensions/subtoken/description.yml`, staging [subtoken](https://github.com/meridian-online/subtoken) for the community-extensions registry.

### Extension

`version: 0.1.0`, `ref: c453860544b53803bb9b830c4818e4f1ad2d4057` — the merge commit that carries the extension's current name, its provenance functions, and the current build contract (`Makefile` and the `extension-ci-tools` submodule).

subtoken turns text into a fixed-width vector inside DuckDB, from a model compiled into the binary: no network call, no API key, no per-row bill.

### Verification

Every SQL statement in `description.yml` was extracted from the YAML with a parser (`scripts/check_description_examples.py` in the source repository, not a regex over the file) and run, in one DuckDB session, against a build packaged from `ref`, with no fixture the descriptor does not create for itself. **Result: 12 statements, 0 errors.** The same check confirmed the entry's version and repository fields agree with the source tree, that it names all 7 functions the built artifact registers, and that each of the 5 platforms the descriptor does not exclude carries a matrix name and an inspection-job entry in the distribution workflow.

This registry's own `Community Extension Build` workflow ran the staged entry on the meridian-online/community-extensions fork (https://github.com/meridian-online/community-extensions/actions/runs/35020622589): every job for a platform the descriptor does not exclude — linux_amd64, linux_arm64, osx_amd64, osx_arm64, windows_amd64 — is green.

### Notes for review

- `excluded_platforms` excludes `wasm_mvp;wasm_eh;wasm_threads` (the bundled model weights make the artifact too large for a wasm target) and `linux_amd64_musl;linux_arm64_musl;windows_amd64_mingw` (opt-in platforms not opted into).
- `requires_toolchains: rust;python3` — python3 backs the metadata-stamp venv the C-API build contract uses.
- This is a first-time submission: no prior entry exists at `extensions/subtoken/`.


lghavami:
Thanks!

PR #2712, opened 2026-09-16
Add mmcif extension

sverhoeven:
Adds [mmcif](https://github.com/i-VRESSE/duckdb-mmcif) extension to read/write tables inside mmCIF files from the [Protein Data Bank](https://www.rcsb.org/).


lghavami:
looks good, thanks!

PR #2715, opened 2026-09-16
ci: publish corresponding source materials with native extensions

kaka11chen:
Community builds currently publish extension binaries without preserving the
source/rebuild materials generated alongside them. For `lerobot`, whose native
build includes FFmpeg, a separate build's source bundle does not establish the
inputs used for the community binary.

Add an opt-in `extension.release_materials` descriptor setting and enable it for
`lerobot`. The build enables its existing packaging target, disables the vcpkg
binary cache so dependency sources are collected, and passes the accompanying
artifact to deployment. Other extensions keep the existing publication path.

The publication helper validates `SHA256SUMS` and the manifest's unsigned-binary
checksum, uploads materials to hash-addressed public storage, then invokes
DuckDB's existing signing/upload script. It downloads each published binary,
checks that only the signature trailer changed, and publishes a checksum-linked
`.duckdb_extension.sources.json` companion and a retained publication receipt.
The helper comes from this workflow repository and never executes artifact code.
Source completeness remains the extension maintainer's responsibility.

**Dependency:** https://github.com/duckdb/extension-ci-tools/pull/412 must land on
`v1.5-variegata` before this workflow can accept `include_release_materials`.
This is a draft for reviewing the opt-in interface and storage/publication
approach; it must not be merged until that dependency and the native packaging
builds are verified. No production uploads were performed by these tests, and
the binary already published from #2706 has not acquired a matching source
companion through this change.

Validation:
- All 15 tests passed locally and in the [publication test workflow](https://github.com/duckdb/community-extensions/actions/runs/35122684598), covering descriptor compatibility, hash/path validation,
  upload ordering, failures before signing, payload tampering/truncation,
  immutable receipts, and filename isolation.
- The integration test used DuckDB 1.5.5's unmodified signing/upload script,
  a disposable RSA key, and local storage in place of AWS. Both latest and
  versioned signed downloads matched their receipts.
- A pre-existing complete `lerobot` source bundle passed the helper's dry run;
  this checks the packaging contract, not its association with the current
  community binary.
- Black and `git diff --check` passed. actionlint reports no new diagnostics
  against the base; the existing deploy workflow still uses `checkout@v3`.

The full build workflows currently cannot start because the referenced shared
workflow does not yet declare `include_release_materials`; the dependency above
adds that input. The publication test workflow runs independently and is green.

Before rollout, the six native community builds must generate matching bundles,
and maintainers must confirm retention and public access for `release_materials/`
in the community bucket. Material uploads precede binary publication; replacement
of a latest binary and its companion is not atomic, so clients must compare the
download checksum. Wasm remains excluded for `lerobot`.


kaka11chen:
Closing this broader CI/deployment proposal while we pursue a smaller, project-managed source distribution approach. We will verify the materials against the community build and provide the download information through the existing REDISTRIBUTION.md link. This closure does not assert that source delivery for the current community binary is already complete.

PR #2718, opened 2026-09-16
zim: v0.9.1

teaguesterling:
Bumps `zim` to **v0.9.1** and sets `ref_next` for the first time.

Compatibility and build-correctness release — **no API or behaviour change** from 0.9.0.

- **DuckDB v2.0 compatibility.** `ConstantExpression(const Value &)` is deleted on `v2.0-cyanoptera`; zim's replacement scan (`SELECT * FROM 'archive.zim'`) used it. It now dispatches through a feature probe for `ConstantExpression::FromValue`. zim builds **and passes its full test suite** on that line, and every PR is now gated on it.
- **The build is pinned to what ships.** The `duckdb` submodule is pinned at exactly the `v1.5.5` tag; it previously sat on an untagged commit 154 past v1.5.4 while the workflow declared v1.5.4 and the registry built v1.5.5. A CI job now asserts the pin equals the shipped tag and matches the registry's own defaults.
- **Golden tests** for byte-exact entry text, nested headings, and a text entry that is not valid UTF-8.

## `ref_next`

`ref` and `ref_next` are the **same commit**, `ce892d5515bcbe5993f2efbe0f22a302a4121e3c`. zim previously carried no `ref_next`, so it has never been built on the 2.0 line; as of v0.9.1 it passes there.

The sha is the **commit**, resolved with `git rev-parse v0.9.1^{commit}`. zim's `v0.9.1` tag is annotated, so a bare `git rev-parse v0.9.1` yields the tag object (`cade6a46`), which the registry cannot build. Noting it because the descriptor cannot show which kind a sha came from. The previous 0.9.0 entry resolved this correctly too.

## Why this bump matters beyond the version number

The registry currently serves zim from **24 August** (`df80e1e`) even though the 0.9.0 entry is correct, because that build failed on a transient `curl (28)` fetch of xapian. So users have been installing a build without the v1.5.5 pin or the v2.0 fix. I'll confirm this release by a clean-extension-directory `INSTALL zim FROM community` once the registry has **built** it, rather than treating the merge as the result.


lghavami:
Thanks!

PR #158, opened 2024-10-21
New Extension: open_prompt

lmangani:
Experimental and rudimental `open_prompt()` extension loosely based on the Motherduck demo at DuckDB Meetup

### Functions
- `open_prompt(prompt::VARCHAR, model::VARCHAR)`
- `set_api_url(url::VARCHAR)`
- `set_api_token(token::VARCHAR)`
- `set_model_name(model::VARCHAR)`

### Settings
```sql
SELECT set_api_token('your_api_key_here');
SELECT set_api_url('http://localhost:11434/v1/chat/completions');
```

### Usage
```sql
D SELECT open_prompt('Write a one-line poem about ducks', 'qwen2.5:0.5b') AS response;
┌────────────────────────────────────────────────┐
│                    response                    │
│                    varchar                     │
├────────────────────────────────────────────────┤
│ Ducks quacking at dawn, swimming in the light. │
└────────────────────────────────────────────────┘
```

### Ollama self-hosted
Test the open_prompt extension using a local or remote Ollama with Completions API

#### CPU only
```
docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
```
#### Nvidia GPU
Install the Nvidia container toolkit. Run Ollama inside a Docker container
```
docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
```

PR #2719, opened 2026-09-16
webbed 2.9.1

teaguesterling:
Bug-fix release. Four reported defects, one external contribution, and the vendored duck_block vocabulary moved to spec 1.4.

Release notes: https://github.com/teaguesterling/duckdb_webbed/releases/tag/v2.9.1

**Three of these change behaviour**, which a patch version does not signal on its own:

- **SAX streaming now preserves CDATA sections, comments and processing instructions** in captured nested fragments (#158). libxml2 discards callbacks that are not registered, so comments and PIs never reached the reader, and `cdataBlock` delegated to `characters`, flattening CDATA framing. Four of five cases diverged from the DOM path. Anything matching streamed fragment output byte-for-byte will see different bytes.
- **Malformed XML fails closed under SAX** instead of recovering into partial data, matching DOM; with `ignore_errors := true` an invalid file contributes no rows rather than a truncated prefix (external contribution).
- **A column mixing boolean and numeric text infers VARCHAR** (#160). Sample types were unified on the value cast lattice, where BOOLEAN casts to INTEGER — but the samples are text, and no integer lexeme spells `true`, so inference produced a type that could not read the data it was inferred from.

Also: `datetime_format` now reaches nested STRUCT fields and LIST elements (#159), declining rather than guessing on genuinely ambiguous dates; and the DuckDB 2.0 fallible contract is now guarded by error class rather than message text (#156).

### `ref_next` added

This descriptor had no `ref_next`. The registry sets `COMMUNITY_EXTENSION_REQUIRE_REF_NEXT: true`, and a descriptor lacking the key is skipped with `Skipping prerelease validation: repo.ref_next is not defined` — which is a **pass**, not a failure.

Measured, rather than assumed: that gate landed between 2026-09-06 20:50Z and 2026-09-07 09:24Z. webbed 2.9.0 was prepared 2026-09-06 05:20Z, so it predates the gate and **did** build on the 2.0 line via `ref`. 2.9.1 is the first webbed release that would have hit the gate — so this key is being added for a forward-looking reason, not to repair a past skip.

Both keys now point at the v2.9.1 tag (`cdd866f`), which builds on both lines. This PR's own `test_against_latest / prepare` log shows zero occurrences of the skip message, with `DUCKDB_VERSION: v2.0-cyanoptera` and the `ref_next` sha resolved. `andium` is left at its prior commit, unchanged.

### Verification

Built against the pinned duckdb submodule (v1.5.5, `d8cdaa33fd` — the repo now asserts in CI that the pin equals the shipped tag). Full suite **4072 assertions in 106 test cases**; the four fixes were also merged together and built as a combination before landing, since two of them touch the same file. `main` CI green across the full platform matrix and the v2.0 line.


teaguesterling:
Correction to the description above, flagged rather than quietly edited.

I originally wrote that webbed's missing `ref_next` meant the prerelease leg had been printing `Skipping prerelease validation` and passing green having verified nothing. **That past-tense claim was wrong**, and I've corrected the description.

Measured from the prepare logs: `COMMUNITY_EXTENSION_REQUIRE_REF_NEXT` is absent from runs on 2026-09-06 07:43Z and 20:50Z, and present (with the skip message) on 2026-09-07 09:24Z — so the gate landed between those. webbed 2.9.0's prepare job started 2026-09-06 05:20Z with no `REQUIRE_REF_NEXT` and no skip message, and `DUCKDB_VERSION: v2.0-cyanoptera`, so it did build on the 2.0 line via `ref`.

The reason for adding `ref_next` is unchanged, just forward-looking rather than remedial: the gate exists now, so 2.9.1 would have been the first webbed release skipped without it. This PR's own `test_against_latest / prepare` log confirms the leg runs — zero skip messages, `ref_next` sha resolved, `DUCKDB_VERSION: v2.0-cyanoptera`.

The commit message carries the same original wording; I've left it rather than force-push and restart the in-flight 2.0 build over a prose fix.

lghavami:
Thanks!

PR #2720, opened 2026-09-16
Bump quack_oauth to v2026.09.16

jrosskopf:
Bumps `quack_oauth` to [`v2026.09.16`](https://github.com/DataZooDE/quack-oauth/releases/tag/v2026.09.16), a correctness release for the authorization object walk. Follows #2478 and #2709; `excluded_platforms` is unchanged.

**Operators running a `policy_table` against DuckDB 1.4.x should take this bump.** On that line a statement wrapped in a `WITH` clause collected no objects at all, and a request with no objects does not engage object-scoped policy rules — so such statements were not evaluated against them.

## What changed

- **Metadata introspection works again.** `v2026.09.15` gated every table function under `fn:<name>`, including the `duckdb_*` family — so `duckdb_tables()` needed an explicit rule while `information_schema.tables`, the same rows, did not. That bought no security and broke routine BI-client introspection. The family is now exempt, matching the catalog views it mirrors, except `duckdb_secrets` and `duckdb_settings`, which carry credential material and have no catalog-view equivalent.

- **CTE handling corrected, both lines.** A CTE alias is no longer treated as an object — it parses as a plain table reference, so `WITH c AS (SELECT * FROM t) SELECT * FROM c` used to surface `main.c`, which does not exist. Separately, DuckDB 1.4.x parses a `WITH` clause into a `CTENode` wrapping the query where 1.5.x attaches it to the enclosing select node; the walk had no case for that node type, so the entire select half of such a statement was skipped on the LTS line. `CTE_NODE` and `RECURSIVE_CTE_NODE` are now handled, and both DuckDB lines produce identical object sets.

- **New `quack_oauth_inspect_sql()`** returns what the authorization layer would gate on — `(action, objects, columns, unsafe, error)` — parsing without executing and consulting no policy. It is also how the two bugs above were found: `InspectSql` links DuckDB's parser and so is unreachable from the pure-logic unit-test binary, leaving the walk without direct coverage until now.

## Verification

CI green on the tag in the source repository: 31 jobs across DuckDB v1.5.5 and v1.4.5 LTS, covering `linux_amd64`, `linux_arm64`, `osx_amd64`, `osx_arm64` and `windows_amd64`, including per-platform smoke tests that install the shipped artifact into a stock DuckDB CLI, load it and call a real function. Test suites verified against both DuckDB lines locally: 1717 unit assertions, 254 SQL assertions on v1.5.5 and 250 on v1.4.5.


lghavami:
Thanks!

PR #2724, opened 2026-09-17
markdown 1.9.3

teaguesterling:
Housekeeping release. **No behaviour change** — one new regression guard and the vendored duck_block vocabulary moved to spec 1.4.

Release notes: https://github.com/teaguesterling/duckdb_markdown/releases/tag/v1.9.3

### The DuckDB 2.0 fallible contract is now guarded by error class, not message text

`duck_block_to_md`, `duck_blocks_to_md` and `duck_blocks_to_sections` all reach a Pandoc depth guard that throws `InvalidInputException`. All three already declared `SetFallible()`, but nothing pinned the error **class** — and DuckDB 2.0 rewrites an unmarked scalar function's execution-error throw into an `INTERNAL Error` that *quotes the original message*. So an expectation matching only the message text passes either way and cannot detect a lost mark; only the `Invalid Input Error: ` prefix discriminates.

That isn't inferred — duck_block_utils ran the experiment: with a mark removed, the 2.0 leg produced the INTERNAL form and the class-prefixed expectation rejected it. No function was found unmarked here; the gap was entirely test-side.

### Vendored duck_block vocabulary at spec 1.4

Vendored at upstream `95a84e6` (duck_block_utils v3.3.0), adding `PREDICATE_REVISION`. Header and documentation only — no source change, and no change to emitted output: markdown's only metadata row is frontmatter, a single level-1 row with no descendants, so 1.4's level-scoped metadata rule drops nothing here.

Upstream has since added `TYPE_DOCUMENT` at the same minor, so this copy reads as one additive constant behind; the drift check reports "OK with news" and exits 0. No re-vendor required for that.

### Descriptor

`version` 1.9.2 → 1.9.3, and `ref`/`ref_next` both move to the v1.9.3 tag (`b923ad8`), which builds on both lines. `ref_next` was already present here, so this is a version and ref bump only. `andium` is left at its prior commit, unchanged.

### Verification

Built against the pinned duckdb submodule (v1.5.5, `d8cdaa33fd`).

- New test: **10 assertions**, passing. The boundary was measured rather than computed — depth 1000 renders on all three surfaces, depth 1001 throws on all three.
- Full suite: **1962 assertions in 56 test cases** — every `.test` file in the repo, verified rather than assumed.
- CI green on the merge commit across the full platform matrix, the DuckDB 2.0 line, vocabulary drift and duck_block conformance.

https://claude.ai/code/session_01XoHU19qnngy1CJnXyX1csz


lghavami:
Thanks!

PR #2725, opened 2026-09-17
yaml: v1.9.1

teaguesterling:
Bumps `yaml` to **v1.9.1** and sets `ref_next` for the first time.

## What changed

- **DuckDB v2.0 compatibility.** `UnaryExecutor::ExecuteWithNulls` was removed upstream in `987ea2c409` and is absent from `v2.0-cyanoptera`. yaml's VARCHAR→YAML cast used it, so the extension had failed to compile on that line since at least 2026-09-13. It now routes through a feature-detecting shim (probing the API, not testing a version macro). The v2.0 legs — `linux_amd64`, `windows_amd64` and `wasm_eh` — all build and pass.
- **Two silent data-loss fixes.** Frontmatter with an unmatched opening delimiter could run on, latch onto a `---` in the body, and promote body text into frontmatter — dropping it from the document. Separately, `yaml_to_json('')` returned a zero-length string from a function declared to return `JSON`, which surfaced downstream as `Malformed JSON at byte 0 of input: input length is 0`; an empty YAML document is now `null`. Both have tests.
- **The build is pinned to what ships.** The `duckdb` submodule is pinned at the `v1.5.5` tag, and CI asserts the pin equals that tag.

No API changes.

## `repo.ref_next`

This descriptor previously had no `ref_next`, so `scripts/build.py` printed `Skipping prerelease validation: repo.ref_next is not defined` and the next line was never built. That is why the compile failure above stayed invisible for days while everything reported green. `ref_next` now points at the same commit as `ref` — v1.9.1 builds on both the stable (`v1.5.5`) and next (`v2.0-cyanoptera`) lines.

`repo.andium` is deliberately left at `0bdc812c` (v1.9.0), so the DuckDB 1.4.5 line is unchanged by this bump.

Release: https://github.com/teaguesterling/duckdb_yaml/releases/tag/v1.9.1


lghavami:
Thanks!

PR #2726, opened 2026-09-17
duck_tails: v1.7.0

teaguesterling:
Bumps `duck_tails` to **v1.7.0** and sets `ref_next` for the first time.

A correctness release: several functions returned confident, plausible, wrong answers and now either return the right one or raise. Flagged explicitly because some calls that previously returned rows now throw.

## Contract changes

- **A NULL repository path raises** instead of resolving the literal string `"NULL"` as a path inside the current repo. Previously `git_log(NULL)` returned 0 rows and `git_branches(NULL)` returned rows from an unrelated repository, with nothing reported.
- **`text_diff_stats()` / `text_diff_lines()` read their arguments.** They previously ignored their input and produced fabricated results; any figure they returned before this release is meaningless rather than merely imprecise.
- **A file stored in Git LFS no longer reads back as its pointer text**, and a byte budget no longer changes what `git_read` reports the file to *be* — classification now runs on the whole blob before the budget is applied.

Also: annotated tags are peeled in `git_parents_each`, `git_log`'s `ref` argument is registered, `_each` functions honour the ref a LATERAL join passes them, paths are matched by component rather than spelling, and repository-discovery failures report their real cause.

## Windows on the v2.0 line

`TEST_EXTENSION_DIR` was built from `DUCKDB_MAJOR/MINOR/PATCH_VERSION`, which are empty when DuckDB is built from a branch rather than a tag — yielding a directory named `v..`. That is legal on Linux and macOS (it silently created a junk folder) and invalid on Windows, where the v2.0-line build failed in this repo's own CMake. It now uses `DUCKDB_NORMALIZED_VERSION`. All Windows legs pass, including the v2.0 `windows_amd64` build.

## `repo.ref_next`

This descriptor previously had no `ref_next`, so `scripts/build.py` printed `Skipping prerelease validation: repo.ref_next is not defined` and the next line was never built. `ref_next` now points at the same commit as `ref` — v1.7.0 builds on both the stable (`v1.5.5`) and next (`v2.0-cyanoptera`) lines, and both are gated in the repo's CI.

Release: https://github.com/teaguesterling/duck_tails/releases/tag/v1.7.0


lghavami:
Thanks!

PR #2728, opened 2026-09-17
Update mssql_ducklake to 0.1.1

VGSML:
A fix release for `mssql_ducklake`, tagged [v0.1.1](https://github.com/hugr-lab/mssql-ducklake/releases/tag/v0.1.1); `ref` is that tag's commit.

A transaction that created or altered two or more tables was refused by SQL Server with `Violation of PRIMARY KEY constraint 'pk_ducklake_schema_versions'`. DuckLake records one row per table whose schema changed in a snapshot, and the primary key this extension puts on that catalog table did not include the table, so the second row of a commit read as a duplicate. Reported by a user running dbt against a Fabric SQL database (hugr-lab/mssql-ducklake#30), fixed in hugr-lab/mssql-ducklake#31.

The key now includes the table id, and the catalog shaping rebuilds a key whose column list an earlier build got wrong, so a catalog created by 0.1.0 is repaired on its next attach rather than keeping the old key. Nothing else changed: same DuckDB v1.5.5 line, same embedded DuckLake pin, same mssql extension requirement (v0.2.5 or newer), same excluded platforms.


lghavami:
Thanks!

VGSML:
Thank you!


пт, 18 сент. 2026 г. в 14:21, Leyli ***@***.***>:

PR #2730, opened 2026-09-18
automation: close old PRs

lghavami:
For PRs labeled ```CI failure``` or ```changes requested```:

* If there is no activity for 23 days, the PR is marked as stale
* If there is no activity for an additional 7 days, the PR is closed

In both cases there is an accompanying message. 

This workflow assumes that unlabeled PRs are fine / still being worked on etc. Based on my experience with this repo thus far this flow makes sense, but other suggestions are welcome!


lghavami:
Closing for now given that Dinos's PR tracker / issue tracker should cover this functionality.

PR #2732, opened 2026-09-18
Add semantic_profile extension

patricktrainer:
Adds `semantic_profile`, a Rust extension that profiles what the **values** in a table mean, rather than what shape they are.

`SUMMARIZE` and similar tools report types, ranges, quantiles and null counts. They cannot see that an `email` column holds `n/a`, that a `weight` column mixes `12 kg` with a bare `26.4`, that a free-text note contains a national ID number, or that `country = 'US'` contradicts `postal_code = 'SW1A 1AA'` in the same row. Those are judgments about meaning, and this extension makes them available as SQL.

Judgments come from [TypeSafe](https://docs.typesafe.ai)'s System One model, which returns calibrated typed answers rather than free text, so every finding carries a probability and the rows it came from.

### How it works

Three steps. Discovery reads what each column actually holds from its values. Selection then decides which of a 20-probe catalogue are worth running on *this* table, which is what makes it adapt rather than run a fixed checklist. Execution sends one request per row carrying every selected probe plus the cross-column checks, so cost is `2 x columns + rows` rather than `columns x rows`.

`sem_cost()` is a dry run that makes no API calls, and `sem_probes()` shows what will be asked before anything is spent.

### Notes for reviewers

- Requires a TypeSafe API key and sends sampled values to that API. This is stated in the extended description; `sem_cost()` reports how much will be sent before any of it is.
- The key is never settable through SQL (it would land in query logs); it comes from the environment or a file.
- Loading the extension does not write to the user's database: the SQL macro layer installs automatically only on an in-memory database, where nothing persists.
- `excluded_platforms` matches what the extension's own CI builds and tests: no wasm target, and the Rust toolchain targets MSVC rather than mingw.
- Built from the `v0.1.0` tag. I verified a clean `--recursive` clone of that tag runs `make configure && make release && make test_release` successfully, with the test suite replaying frozen API responses offline so it needs no key.


sebastiaan-dev:
Thanks!

PR #2734, opened 2026-09-19
Update dicom extension to v0.7.1

nmontesg:
This update fixes the `read_dicom` function for files with SpecificCharacterSet "ISO 2022 IR 100". Now conversion to UTF-8 for them is handled by boost.Locale.


sebastiaan-dev:
Hi @nmontesg, can you change the `ref` tag to be a commit SHA?

PR #2737, opened 2026-09-20
erpl_web: bump to v2026.09.20

jrosskopf:
Version bump for `erpl_web` to [v2026.09.20](https://github.com/DataZooDE/erpl-web/releases/tag/v2026.09.20).

**This release contains security fixes**, so it would be good to have it out: a bearer token written to the trace file on disk, a client secret POSTed to an unvalidated host, OData quote injection returning a different record than the one asked for, predictable PKCE verifiers and CSRF state (mt19937 seeded with 32 bits, now RAND_bytes), and an XSS sink on the loopback origin that receives authorization codes.

Also fixes several silent wrong answers: catalog listings that presented truncated results as complete, and a describe that returned the literal string `error` as a record's name rather than failing.

No changes to the extension's build, toolchain, or platform exclusions.


sebastiaan-dev:
Thanks!

PR #2738, opened 2026-09-20
bump to 0.1.6

paulmupeters:
Fixed manual SharePoint token authentication so the TOKEN secret parameter is accepted and persisted by the token provider.


sebastiaan-dev:
Thanks!

PR #2739, opened 2026-09-20
gpudb: update to v0.7.0

singhpratech:
Updates `gpudb` from v0.6.0 to v0.7.0 (`ref` → `b712f4b7624b83873ca9c221f70a4a9b5ce5f3f7`, tag `v0.7.0`).

What's new in this version:
- Plain DuckDB SQL on the GPU. Through the project's `gpudb` shell or `gpudb.connect()` (`pip install duckdb-gpudb`), ordinary statements — `GROUP BY`, `WHERE`, `HAVING`, `ORDER BY … LIMIT k`, joins, `EXISTS` / `IN` subqueries, CTEs, views — are answered on the device when that is measured faster and by DuckDB otherwise, with identical rows. `LOAD gpudb` in any client keeps giving the explicit `gpu_*` functions, as before.
- Exact `GROUP BY`, filters, global aggregates and joins on both backends (Apple Silicon Metal and NVIDIA CUDA): integer and `DECIMAL` aggregates are bit-exact with 128-bit sums on the device; `avg` is finalised the way DuckDB finalises it; a pushed top-k whose ordering values tie is handed back to DuckDB.
- 65 `gpu_*` functions (38 in v0.6.0). None were removed and no signature changed.
- Still the stable C API only (C_STRUCT ABI): no libduckdb link, no DuckDB C++ headers.

The description is rewritten for this version; the example and the platform notes describe what this registry's build gives, including that a Linux binary from here may be CPU-only and needs `libgomp.so.1` at load time.

Results are compared with native DuckDB before any time is counted; the numbers, including the shapes where the GPU loses, are in the repo's BENCHMARK.md and in the extended description here.

Fork validation run of this branch — the same workflow, building and testing the tagged commit on all four platforms, green: https://github.com/singhpratech/community-extensions/actions/runs/35537293620

Thanks for the sweep as always.


sebastiaan-dev:
Thanks!

PR #2740, opened 2026-09-21
Add jev extension

judoaseeta:
`jev` lets you filter, rank and classify rows with plain-language conditions.

```sql
INSTALL jev FROM community;
LOAD jev;
SET jev_api_key = '...';

SELECT * FROM tickets WHERE jev(tickets, 'the customer wants a refund');

SELECT body, jev_prob(tickets, 'the customer is angry') AS p FROM tickets ORDER BY p DESC LIMIT 20;

SELECT jev_choice(tickets, 'which team should handle this?',
                  ['billing', 'technical', 'sales']) AS team, count(*)
FROM tickets GROUP BY 1;
```

Each row is judged by [TypeSafe's Jev](https://docs.typesafe.ai), a model that returns calibrated
probabilities instead of generated text, so there is no index, no embeddings and no vector column.
`jev()` is an ordinary boolean function, so it composes with the rest of SQL.

This is a port of [pg-jev](https://github.com/realZachi/pg-jev) ([pgjev.com](https://pgjev.com)) to
DuckDB, and speaks the same API. Where pg-jev needs a read-ahead that streams the table in physical
order — Postgres hands a function one row at a time — DuckDB already delivers a vector of up to 2048
rows, so rows go out `jev_batch_size` (20) per request over a process-wide pool of `jev_concurrency`
(16) kept-alive connections. Answers are cached by row content, and rows that a cheaper predicate in
the same `WHERE` rejects first are never judged.

### Notes for review

- **Network**: the extension calls `https://api.typesafe.ai/v1/systemone` (overridable with
  `jev_api_url`) and needs an API key, from `jev_api_key` or `TYPESAFE_API_KEY`. Row contents are sent
  to that API; the README says so plainly. `jev_max_rows_per_statement` and
  `jev_max_chars_per_statement` are there as spend guards.
- **Platforms**: wasm is excluded — the extension needs sockets and a thread pool. Everything else in
  the matrix (linux_amd64, linux_arm64, osx_amd64, osx_arm64, windows_amd64, windows_amd64_mingw) is
  green against v1.5.5 in the extension's own CI:
  https://github.com/judoaseeta/duckdb-jev/actions/runs/35548289036
- **Tests**: the regression suite runs against a deterministic mock of the API
  (`test/mock_api.py`), so `make test` never reaches the live service and needs no key. The tests that
  do need the mock are gated behind an environment variable and skip without it.
- **Dependencies**: OpenSSL through vcpkg; cpp-httplib and nlohmann/json are vendored under
  `third_party/`, with httplib renamed to `duckdb_jev_httplib` so it cannot collide with the copy
  DuckDB ships.


rustyconover:
You could make it work on wasm if you call HTTPUtils, and Jev offers the right CORS headers.

PR #2749, opened 2026-09-21
Update markdown to v1.9.5

teaguesterling:
Updates markdown to **v1.9.5** (`437e204`).

Retargeted from v1.9.4, which should not ship: it carries a crash that prevents its own build from publishing.

## Why this is not v1.9.4

`ExtractPandocTextFromVal`, the Pandoc walker, had two defects:

- its **array arm never incremented the depth counter**, so nested arrays added stack frames the guard could not see — no value of `MAX_PANDOC_DEPTH` bounded them (100000 nested arrays segfaulted on an 8 MB stack while the same depth in objects threw cleanly)
- the cap of **1000 was larger than a 512 KB macOS worker thread can carry** at ~275 bytes/frame, so the stack died around depth 900, before the guard at 1001 could fire

That is what stopped **v1.9.3** from publishing after #2724 merged on 2026-09-17: the post-merge macOS arm64 build died with `make: *** [test_release_internal] Bus error: 10`, and a failed build skips the `deploy` job, so the merge completed and `INSTALL markdown FROM community` silently kept serving v1.9.2.

Worth flagging for anyone reading this pattern: **green PR checks were not evidence against it.** The failure is marginal, not deterministic — it depends on how much stack the harness above the walker already used. v1.9.3's registry PR passed its macOS legs, this PR passed its macOS arm64 leg in 10m30s at v1.9.4, and the extension's own CI passed macOS on all of those commits. Only the post-merge build failed.

Fixed in teaguesterling/duckdb_markdown#76 (issue #75): every recursion now counts, and the cap is 200 — about 55 KB of walker, ~9× margin inside a 512 KB stack, still allowing ~100 levels of nested emphasis. Verified green on the full suite and, critically, on the full suite under `ulimit -s 512`, the condition that actually reproduced the crash.

## `ref_next` is also corrected here

This PR previously carried `ref: cd50cca` (v1.9.4) with `ref_next: b923ad8` — still **v1.9.3's** commit. The prerelease leg would have validated the previous release against DuckDB 2.0 while the descriptor claimed 1.9.4. A missing `ref_next` announces itself ("Skipping prerelease validation"); a stale one builds a real tree and passes green. Both now point at `437e204`.

## Contents

v1.9.5 delivers its own crash fix plus the two releases that never reached users: the DuckDB 2.0 fallible-contract guard (v1.9.3) and `duckdb_functions()` discoverability with descriptions, parameter names and examples (v1.9.4).

Release: https://github.com/teaguesterling/duckdb_markdown/releases/tag/v1.9.5

https://claude.ai/code/session_01XoHU19qnngy1CJnXyX1csz


teaguesterling:
Converting this to draft — please don't merge it yet. Marking it rather than leaving it mergeable because merging it would burn a third release, and the reason isn't visible from this PR's own checks.

**`cd50cca` (v1.9.4) carries a crash that prevents the build from publishing.** `ExtractPandocTextFromVal`'s array arm recursed without incrementing its depth counter, and the depth cap (1000) was larger than a 512 KB macOS worker thread can hold — the walker costs ~275 bytes per frame, so the stack died around depth 900, before the guard at 1001 could fire.

That is exactly what happened to **v1.9.3**, whose descriptor merged here as #2724 on 2026-09-17: the post-merge macOS arm64 build died with `make: *** [test_release_internal] Bus error: 10` (job `105198327838`). A failed build skips the `deploy` job, so the merge looked fine and `INSTALL markdown FROM community` silently kept serving v1.9.2.

Worth flagging for anyone hitting something similar: **this PR's green checks are not evidence against it.** The failure is marginal, not deterministic — it depends on how much stack the harness above the walker has already used. v1.9.3's PR passed its macOS legs too, and then its merge build crashed. The extension repo's own CI passes the macOS legs on these commits as well.

The fix is in flight at teaguesterling/duckdb_markdown#75. Once it lands I'll retarget this PR to v1.9.5 and mark it ready, rather than opening a new one.

https://claude.ai/code/session_01XoHU19qnngy1CJnXyX1csz

teaguesterling:
Lifting my earlier "please don't merge this yet" — this PR now points at **v1.9.5** (`437e204`) and is ready.

What changed since that note:

- **The crash is fixed** (teaguesterling/duckdb_markdown#76, issue #75). Every recursion in the Pandoc walker now counts toward the depth guard — the array arm previously did not, so nested arrays were unbounded at any cap — and the cap moved from 1000 to 200, which fits inside a 512 KB macOS worker stack with ~9× margin instead of needing ~275 KB and dying first.
- **`ref_next` is corrected.** It was still `b923ad8` (v1.9.3) while `ref` had moved to v1.9.4, so the DuckDB 2.0 prerelease leg would have validated the *previous* release while the descriptor claimed the new one. Both are now `437e204`.
- Verified on the full suite and under `ulimit -s 512`, which is the condition that actually reproduced the macOS crash; a plain run on an 8 MB stack never does.

Thanks for the patience — drafting it was the right call, since the version it carried would have failed to publish for the third release running.

sebastiaan-dev:
Thanks!

PR #2755, opened 2026-09-21
Update panduck to v0.5.5

teaguesterling:
Updates `panduck` to `v0.5.4` (commit `ea891eddedf65fd68fa5ce4ce431fbe133f1aa0a`):
- Adds `FunctionDescription` documentation metadata (parameter names, description, examples, categories)
- Adds `OnCreateConflict::ALTER_ON_CONFLICT` to function registrations for discoverability via `duckdb_functions()`


teaguesterling:
Pushed two corrections to this branch, and tagged the release it names.

**`ref_next` was stale.** It pointed at `4dc6985` (the v0.5.3 tag) while `ref` moved to `ea891ed`. With those disagreeing, `test_against_latest` builds the *previous* release against the next DuckDB while the descriptor ships a newer one — a green that measures the wrong commit. That job has caught two real defects for panduck before (a v2.0 source break, and a missing `catalog.hpp` include), so it is worth having it point at what is actually being published. Both fields now name the v0.5.4 commit.

**`v0.5.4` now exists as a tag and a release.** When this PR was opened, `ref` named `ea891ed` but no tag or release pointed at it — tags stopped at v0.5.3. It is now [v0.5.4](https://github.com/teaguesterling/duckdb_panduck/releases/tag/v0.5.4), annotated, on that exact commit.

**Also re-measured** the `extended_description`'s assertion count, 2931 → 2956; the CI job count it claims (eleven) is still correct.

Verified on `ea891ed` before tagging: full CI matrix green (28 non-canary jobs), suite passes 2,956 assertions in 30 files, no reader drops a word pandoc can see, and every skipped test file is declared.

Issue #91, opened 2024-09-03
List of extensions in the website is currently not auto-generated


carlopi:
https://community-extensions.duckdb.org/list_of_extensions.html needs to be also bumped automatically.

carlopi:
Fixed by #92 + #93

PR #2759, opened 2026-09-21
duckdb_mcp: pin ref_next so the v2.0 leg is actually built

teaguesterling:
No version change. This adds `repo.ref_next` to the `duckdb_mcp` descriptor.

`ref_next` is set to the same commit as `ref` (v2.3.0).


sebastiaan-dev:
Hi @teaguesterling, this has a conflict as well.

teaguesterling:
Superseded by the v2.3.2 update (ref == ref_next). This PR pinned ref_next to main — validating that main compiles on v2.0 — but left the shipped v2.3.1 (which predates the #91 v2.0 fix) unvalidated. v2.3.2 folds the fix into the shipped ref and validates it via test_against_latest.

PR #2722, opened 2026-09-17
sitting_duck 1.15.1

teaguesterling:
Bumps `sitting_duck` from 1.11.0 to **1.15.1**, with `ref` and `ref_next` both at `648b570`.

### Why `648b570` and not the `v1.15.1` tag

`648b570` is 3 commits past tag `v1.15.1` (`b52cf325`), deliberately — those 3 commits are what make the extension compile on the v2.0 line:

```
648b570 Merge pull request #175 from teaguesterling/fix/doc-helper-version-agnostic
8568563 fix(doc_helper): see through a set's element pointers too
114920e fix(doc_helper): read argument types through both DuckDB APIs
```

Pinning `b52cf325` (the literal v1.15.1 tag) fails the prerelease leg:

```
src/include/function_doc_helper.hpp:15: error: 'class duckdb::ScalarFunction' has no member named 'arguments'
  :35  ...shared_ptr<const duckdb::ScalarFunction>...     has no member named 'arguments'
  :76  ...shared_ptr<const duckdb::TableFunction, true>... has no member named 'arguments'
```

`v2.0-cyanoptera` keeps `arguments` public on `SimpleNamedParameterFunction` (`TableFunction`'s base) but routes `ScalarFunction` through `FunctionSignature`, where it sits behind `GetArguments()`. The set errors are a second, separate change: `ScalarFunctionSet::functions` holds values on v1.5.4 and `shared_ptr<const ScalarFunction>` on the v2.0 line.

### Changes since 1.11.0

`ast_select` CSS-selector correctness batch:
- `:scope` / `:in-scope` split (`:scope` = "is a scope boundary"; containment is `:in-scope`)
- Bare type selectors are **exact**, not prefix; `[type^=]`/`[type$=]`/`[type*=]` added
- `[file]`/`[line]` attribute filters; `ast_selector_for` ("copy selector")
- Malformed selectors raise instead of silently over- or under-matching
- `:has`/adjacency fixes; `IS_CONSTITUENT` for construct sub-parts
- `:called-by`/`:calls` are direct and lambda-aware, bounded via `scope.function` (fixes an OOM on large tables)

In 1.15.x:
- `ast_unparse`: pluggable rules-based AST unparser, with default rules across 27 languages
- Functions are discoverable from `duckdb_functions()`
- DuckDB v2.0 compatibility: `ConstantExpression` → `Literal`, plus the doc-helper argument-type fix above

### Supersedes

[...]

PR #2764, opened 2026-09-21
sitting_duck: pin ref_next so the v2.0 leg is actually built

teaguesterling:
No version change. This adds `repo.ref_next` to the `sitting_duck` descriptor.

`ref_next` is set to the same commit as `ref` (v1.11.0).


teaguesterling:
Closing: superseded by #2722, which is the better change.

This PR only added `repo.ref_next` while leaving `version` at 1.11.0 and `ref` at `b8c06a84`. It did its job — it made the prerelease leg build instead of printing "Skipping prerelease validation" and passing green — and what it surfaced was real:

```
src/include/duckdb_compat.hpp:351: error: 'const class duckdb::ConstantExpression'
  has no member named 'GetValue'
```

I then repointed `ref_next` at `b52cf325` (= tag `v1.15.1`), which cleared that and exposed the next one:

```
src/include/function_doc_helper.hpp:15: error: 'class duckdb::ScalarFunction' has no member named 'arguments'
  :35  ...shared_ptr<const duckdb::ScalarFunction>...    has no member named 'arguments'
  :76  ...shared_ptr<const duckdb::TableFunction, true>... has no member named 'arguments'
```

v2.0-cyanoptera keeps `arguments` public on `SimpleNamedParameterFunction` (`TableFunction`'s base) but routes `ScalarFunction` through `FunctionSignature`, where it sits behind `GetArguments()`; the set errors are a separate change, since `ScalarFunctionSet::functions` holds values on v1.5.4 and `shared_ptr<const ScalarFunction>` on the v2.0 line.

That is fixed upstream in teaguesterling/sitting_duck#175, merged 2026-09-22, and #2722 now pins `ref` and `ref_next` at `648b570` — main's tip, which contains it. So #2722 delivers a correct version bump *and* a working prerelease pin, where this PR would pin a commit known not to compile and would conflict with it on the same descriptor.

No action needed here. Verification for sitting_duck on the v2.0 line belongs to #2722's `test_against_latest` leg.

PR #2765, opened 2026-09-21
urlpattern: pin ref_next so the v2.0 leg is actually built

teaguesterling:
No version change. This adds `repo.ref_next` to the `urlpattern` descriptor.

`ref_next` is set to the same commit as `ref` (v0.4.1).

This descriptor also pinned `ref: refs/tags/v0.4.1`. That is replaced with the raw commit it resolves to (`33d5e897`), since a catalog ref must be immutable and a tag can be moved. The extension's contents are unchanged.


teaguesterling:
Repointed `ref_next` off `ref` and onto current `main`.

The original pin set `ref_next = ref`, and that did its job — it surfaced a real break rather than the silent skip this PR set out to fix:

```
src/urlpattern_extension.cpp:1821: error: 'class duckdb::ScalarFunction'
  has no member named 'varargs'; did you mean 'GetVarArgs'?
src/urlpattern_extension.cpp:1822: error: 'class duckdb::ScalarFunction'
  has no member named 'null_handling'; did you mean 'GetNullHandling'?
```

v0.4.1 was tagged 2026-07-19 and assigns those as public fields. `v2.0-cyanoptera` moved both into `FunctionSignature` behind `Get/SetVarArgs` and `Get/SetNullHandling`, with `varargs` now private.

That is already fixed upstream — commit `25284e3` (2026-09-05) routes them through feature-detecting helpers in `src/include/duckdb_compat.hpp`, probing for the API rather than testing a version macro. So `ref_next` now points at `0b7dae3e` (urlpattern `main`, which contains that commit) while `ref` keeps shipping released v0.4.1 on the stable line.

`ref_next` is not required to equal `ref` — `duck_lineage`, `http_stats` and `elasticsearch` already carry a separate v2.0 pin. No version change in this PR.

PR #2767, opened 2026-09-22
duck_block_utils 3.4.0

teaguesterling:
Bumps `duck_block_utils` to **3.4.0** (release [v3.4.0](https://github.com/teaguesterling/duckdb_duck_block_utils/releases/tag/v3.4.0), commit `39941a7`).

`ref` and `ref_next` both point at the tagged commit, resolved through `^{commit}`. The `andium` pin is deliberately left at `125662df` — that tree targets DuckDB v1.5.5 and carries no v1.4.x support, as the existing comment in the descriptor records.

## What's in it

**The explicit document root.** Level 0 becomes the optional document root: a `kind='block'`, `element_type='document'` row standing for the document itself. Everything else still starts at 1, so every document valid before this release is valid after it. Several roots are permitted in one relation (a *file* may hold more than one root document; a *document* has exactly one root), a root carries no content — file-scoped facts go in `attributes` — and `metadata`/`value` rows and negative levels remain rejected at level 0. Adds one constant, `TYPE_DOCUMENT`.

`SPEC_VERSION` stays `1.4` and `PREDICATE_REVISION` stays `1.3`: this is additive, and no consumer is forced to re-vendor.

**Fixes.** The Pandoc writers no longer fabricate prose for a document root — the shared writer's terminal fallback turned any unmapped `element_type` into a paragraph, so an empty root became a blank `Para` and a content-carrying root had its text emitted as document prose. The consumer-alignment checker now reads integer constants, enforces major equality plus a minor floor, and compares provenance byte-exactly.

**Also.** Every scalar that can raise an execution error has a test pinning its error class, and functions are discoverable from `duckdb_functions()` with descriptions and examples.

## Verified

Format-check clean, build clean, full suite green (1356 assertions, 18 files) on the pinned DuckDB `d8cdaa33` (v1.5.5); CI green on the release commit (32 success, 1 skipped).


sebastiaan-dev:
Thanks!

PR #2768, opened 2026-09-22
Add pql extension

waelbenamara:
Adds PQL, predictive queries as SQL for DuckDB: `TRAIN MODEL`, `PREDICT`, `BACKTEST MODEL`, `EXPLAIN MODEL` and `DROP MODEL`. A model is trained directly on the tables in the catalog, following foreign keys to build features from related tables, with forecast targets labelled at anchor times so nothing after the anchor leaks into the features.

Built from the extension template against DuckDB v1.5.5 with extension-ci-tools v1.5-variegata. No external dependencies (no vcpkg packages), C++17 only. The distribution pipeline passes on all platforms in the source repository.

Source: https://github.com/Guepard-Corp/duckdb-pql


waelbenamara:
Updated: `ref_next` is now pinned, so the prerelease leg builds instead of printing `Skipping prerelease validation: repo.ref_next is not defined` and passing green having built nothing.

PQL is a parser extension, so one commit cannot satisfy both APIs: between v1.5.5 and v2.0 the parse hook changed from receiving the statement as a string to receiving a token stream it reports back how much of it consumed, result column names changed from `string` to `Identifier`, and the cancellation and cast APIs moved. So the two refs point at two branches of the same repository:

| descriptor | branch | builds against |
|---|---|---|
| `ref` | `main` | v1.5.5 |
| `ref_next` | `next` | `v2.0-cyanoptera` |

Only `src/pql_extension.cpp` differs between them; the language itself, `src/include/pql/pql.hpp`, is byte-identical on both and includes no DuckDB header. Both branches build clean on every platform in the extension's own CI and pass its full suite (229 SQL assertions plus six standalone suites, the fuzzer and degenerate-data harness under ASan/UBSan). The intent to track v2.0 is documented in the README and at the top of the binding on both branches, so it does not get lost.

`ref` also moves forward slightly to pick up descriptions for `pql_exec` and `pql_models`, so the generated page lists them with a description and an example rather than NULL.

waelbenamara:
The workflow runs on `96d5fe6` are sitting in `action_required`: the `ref_next` update re-triggered the first-time-contributor approval gate after the earlier runs on `c1074a7` had passed (Build, and the prerelease leg, both green there). Could someone approve the run when convenient? Happy to answer anything in the meantime.

